PR #63, opened 2015-01-13
Logary.Adapters.Suave targets AnyCPU

jamesjrg:
I can't build it to test as I don't have the Ruby build system


haf:
Could you send a PR with a diff that's human readable? I prefer to merge things I can understand without doing a lot of reassurance, especially since you say you haven't tested building it yourself.

The DotSettings could also be put in .gitignore.

PR #71, opened 2015-05-07
Changes message check to null to match exception thrown in LoggerExtensions.

jbradford:
The check in LoggerExtensions for message is String IsNullOrWhitespace, but the exception thrown is NullArgument, I have changed the check to null equality to match that of the .net LoggerExtensions.


haf:
@jbradford Thank you! That's a valid change.

PR #72, opened 2015-05-11
Use single = for comparison

neoeinstein:


haf:
@neoeinstein thank you! =)

PR #73, opened 2015-05-11
Graphite target using dp typename instead of path

neoeinstein:


haf:
This is awesome, thank you. I know the graphite target didn't work for this reason.

What would you say, if I asked; could you write some tests for this, please?

neoeinstein:
I would say that I need to figure out a good way to do this. I'll get on adding some tests both for writing the path as well as the sanitization.

Also, I think there is some need in there, somewhere, for adding in some additional "namespacing", especially for graphite instances that are the target of may data sources. Prepending something like `YourChosenNameSpace.machinename.` to the path would make them easier to distinguish. Add that to the configuration of the target?

haf:
Yes, on a target level namespacing should be just another setting.

Remember you have the 'service name' in the input to the target, too. This was introduced because you'll need to see what service is causing the error.

When I was writing the code I assumed you wouldn't care about the machine, because you would scale out and then machines wouldn't matter - only the services.

src/Logary/Targets_Graphite.fs
```diff
@@ -46,6 +46,15 @@ let private tryDispose (item : 'a option) =
         | :? IDisposable as disposable -> try disposable.Dispose() with _ -> ()
         | _ -> ()
 
+let private sanitizePath (DP paths) =
```
haf: You could mark it internal and test it.

haf:
Alright, let's merge and release, you can write tests afterwards, given that this target doesn't work at all at the moment.

PR #89, opened 2015-10-24
logstash: allow default tags for logstash target

coolya:
added a defaultTags parameter to the create function and append these
tags to all messages logged in addition to the supplied tags by the user.

I have not added tests yet because haven't replicated your logstash testing environment yet. I first wanted to check if there is any chance to get this merged?


haf:
@coolya Welcome as a contributor to Logary. It would be great to get more PRs like this from you!

For v4: /cc @paavohuhtala – so we can bring it up into the next version.

haf:
If you'd send another PR with test, that would be great. But in general, when it's only about doing a `string list` concatenation, I think the type system would catch the error.

PR #91, opened 2015-11-21
Incorrect syntax by first run

vilinski:
In the table initialization the description this [string](https://github.com/logary/logary/blob/9057f8be5c64f4af81515d6d22a42a054c5200b3/src/targets/Logary.Targets.DB.Migrations/Migrations.fs#L20) causes an exception:

    ....
    .WithColumnDescription("Where's the metric taken from; see graphite type paths")
    ....                         ^

Incorrect syntax bug:
> Error 102: Incorrect syntax near 's'.
> Unclosed quotation mark after the character string ''.

First thought it is a plain string as used in later sql statement. Now seems to be a Fluentmigrator bug. Therefore quick fix for now.


haf:
Good stuff. Thanks!

Where's the corresponding Fluent bug report?

vilinski:
see my PR here https://github.com/schambers/fluentmigrator/pull/679

haf:
@vilinski Great! Thanks for your contribution!

haf:
It's also worth noting that for long-term usage, you may consider to plan to give a hand for moving this target up to v4 (see PR #90)

Issue #81, opened 2015-08-11
Remove FSharp.Actors


haf:
Remove it in favour of Hopac:

 - Update the Registry to match
 - Update the built-in Targets
 - Update all the other Targets

paavohuhtala:
This is being worked on in my branch [feature/hopac](https://github.com/paavohuhtala/logary/tree/feature/hopac).

haf:
Fixed in #90

PR #92, opened 2015-11-25
Replace FSharp.Actor with Hopac progress (4.0)

paavohuhtala:
Progress tracking for #81.

- [x] Core library & tests
- [x] Named jobs
- [x] Supervisor
- [x] Non-core targets
- [x] Performance benchmarking / optimization


haf-raysearchlabs:
@paavohuhtala What state is the metrics in currently? I might have the opportunity to test it out a bit.

.semver
```diff
@@ -1,6 +1,6 @@
 ---
-:major: 3
-:minor: 3
-:patch: 4
-:special: rc
+:major: 4
+:minor: 0
+:patch: 0
+:special: ''
```
haf-raysearchlabs: Should be a pre release

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    albacore (2.4.3)
+    albacore (2.5.1)
```
haf-raysearchlabs: We can upgrade this, also see https://github.com/Albacore/albacore/issues/191

examples/heka/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM gliderlabs/alpine:edge
```
haf-raysearchlabs: Should use https://github.com/fsprojects/docker-fsharp

examples/heka/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+heka:
```
haf-raysearchlabs: This is a good sample to see whether the refactor works

paket.lock
```diff
@@ -20,7 +20,17 @@ NUGET
       FluentMigrator (>= 1.4.0.0)
     FParsec (1.0.1)
     FsCheck (0.9.2.0)
+    FSharp.Control.AsyncSeq (2.0.1)
     FSharp.Core (4.0.0.1)
+    FSharpx.Async (1.12.0)
```
haf-raysearchlabs: This can be removed now that we're using Hopac

paket.lock
```diff
@@ -20,7 +20,17 @@ NUGET
       FluentMigrator (>= 1.4.0.0)
     FParsec (1.0.1)
     FsCheck (0.9.2.0)
+    FSharp.Control.AsyncSeq (2.0.1)
```
haf-raysearchlabs: This can be removed now that we're using Hopac

src/Core.sln
```diff
@@ -0,0 +1,29 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2012
+Project("{f2a71f9b-5d33-465a-a702-920d77279786}") = "FSharp.Actor-logary", "FSharp.Actor-logary\src\FSharp.Actor\FSharp.Actor-logary.fsproj", "{AA4BC25B-04F7-4B39-BF28-0B0058193ED9}"
```
haf-raysearchlabs: Can be removed from the solution

src/Heka.userprefs
```diff
@@ -0,0 +1,16 @@
+﻿<Properties StartupItem="FSharp.Actor-logary/src/FSharp.Actor/FSharp.Actor-logary.fsproj">
```
haf-raysearchlabs: Should be gitignored

src/Logary.CSharp/LoggerExtensions.cs
```diff
@@ -133,43 +136,41 @@ public static T TimePath<T>(this Logger logger, string path, Func<T> f, LogLevel
         /// <param name="args">Arguments to the format string</param>
         public static void LogFormat(this Logger logger, LogLevel level, string formatStringMessage, params object[] args)
         {
-            logger.Log<object>(string.Format(formatStringMessage, args), level, null);
+            // TODO/CONSIDER: Should most of this be moved to the core library?
```
haf-raysearchlabs: Yes

src/Logary/DataModel.fs
```diff
@@ -0,0 +1,774 @@
...
+  | Int64 of int64
+  | BigInt of bigint
+  | Binary of byte [] * ContentType
+  | Fraction of int64 * int64
+  | Object of Map<string, Value> // NOTE: moved from ComplexValue
+  | Array of Value list // NOTE: moved from ComplexValue
+
+  (* Isomorphisms *)
+
+  static member StringPIso : PIso<Value, string> =
```
haf-raysearchlabs: All of these can be updated to the newer `String_` syntax or Chiron/Aether.

src/Logary/DataModel.fs
```diff
@@ -0,0 +1,774 @@
...
+
+type Value =
+  | String of string
+  | Bool of bool
+  | Float of decimal
+  | Int64 of int64
+  | BigInt of bigint
+  | Binary of byte [] * ContentType
+  | Fraction of int64 * int64
+  | Object of Map<string, Value> // NOTE: moved from ComplexValue
```
haf-raysearchlabs: Comments can be removed.

src/Logary/DataModel.fs
```diff
@@ -0,0 +1,774 @@
...
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Value =
+    open Logary.Internals
+    open System.Collections.Generic
+
+    let rec fromObject : obj -> Value = function
+
+    // Built-in types
+    | :? bool as b    -> Bool b
```
haf-raysearchlabs: Coding style: indent one level, to show it's part of method `fromObject`.

src/Logary/HealthCheck.fs
```diff
@@ -42,96 +38,89 @@ module HealthCheck =
...
   [<CompiledName "TryGetDescription">]
   let tryGetDesc m =
-    m.m_data |> Map.tryFind Description |> Option.fold (fun s t -> t :?> string) ""
-
-  /// Tries to get an exception from the measure
-  [<CompiledName "TryGetException">]
-  let tryGetExn m =
-    m.m_data |> Map.tryFind Exception |> Option.map (fun x -> x :?> exn)
+    match Message.tryGetField Description m with
+    | (Some (Field (String s, _))) ->  s
```
haf-raysearchlabs: Should not return `""`, instead return Option like it was before.

src/Logary/HealthCheck.fs
```diff
@@ -42,96 +38,89 @@ module HealthCheck =
...
-    (fun (inbox : IActor<_>) ->
-      let rec running state = async {
-        let! msg, mopts = inbox.Receive()
-        match msg with
-        | GetResult chan ->
+  let private mkFromFunction (fn : unit -> Job<HealthCheckResult>) =
+    let ch = { reqCh = Ch.Now.create () }
+
+    let rec running _ = job {
+        let! req = Ch.take ch.reqCh
```
haf-raysearchlabs: This is fine for requests, because they won't ever overwhelm the system. For input, however, we should use bounded message buffers.

src/Logary/HealthCheck.fs
```diff
@@ -42,96 +38,89 @@ module HealthCheck =
...
+    Job.start fjob |> ignore
+
     { new HealthCheck with
         member x.Name = name
         member x.GetValue () =
-          a
-          |> Actor.reqReply GetResult Infinite
-          |> Async.RunSynchronously
+          (job {
+            // TODO / PERF?: This channel could be reused or replaced with an IVar
```
haf-raysearchlabs: Yes, because we only need a single result.

src/Logary/Internals.fs
```diff
@@ -1,22 +1,38 @@
...
     sprintf "%s/%s" ActorRootNs subcomponent
-    |> fun str -> Actor.Options.Create str
+    |> NamedJob.create
 
 module internal Seq =
+  open Hopac
+
   let all f s = Seq.fold (fun acc t -> acc && f t) true s
   let any f s = Seq.fold (fun acc t -> acc || f t) false s
   let pmap (f : _ -> Async<_>) s = s |> Seq.map f |> Async.Parallel
```
haf-raysearchlabs: Can be removed?

src/Logary/Internals.fs
```diff
@@ -205,6 +221,43 @@ module internal Set =
     | (s : _ Set) when s.Count = 0 -> Some EmptySet
     | _ -> None
 
+type Timeout =
```
haf-raysearchlabs: We probably don't need this anymore because it's built into Hopac in a nice way. Don't you think?

src/Logary/Internals.fs
```diff
@@ -205,6 +221,43 @@ module internal Set =
...
+  let withTimeout timeout j =
+    match timeout with
+    | Infinite -> Job.map Success j
+    | Timeout ts -> job {
+      let! isDone = Promise.start j
+      return! (timeOut ts ^->. TimedOut) <|> (Promise.read isDone ^-> Success)
+    }
+
+  /// Returns a wrapped job and a function for checking the status of the job.
+  let watch (j : Job<_>) =
```
haf-raysearchlabs: Scary

src/Logary/Internals_Logger.fs
```diff
@@ -17,59 +18,40 @@ type NullLogger() =
...
-        member x.Level =
-          x.level
+            logger.Log (fMsg ()) // delegate down
+
+        member x.Log msg =
+          for accept, t in x.targets do
+            if accept msg then
+              Job.Global.start (Ch.give t.reqCh (Log msg))
+
+        // TODO / CONSIDER: This function is just an alias for .Log.
```
haf-raysearchlabs: Let's remove it and create a module for timing functions.

src/Logary/Internals_Logger.fs
```diff
@@ -89,16 +71,13 @@ type InternalLogger =
...
-          x.trgs |> List.iter (flip Target.sendLogLine line)
-      with _ -> ()
-    member x.Measure (m : Measure) =
-      try
-        if m.m_level >= x.lvl then
-          x.trgs |> List.iter (flip Target.sendMeasure m)
+        if msg.level >= x.lvl then
+          x.trgs |> List.iter (Target.send msg >> ignore)
       with _ -> ()
+    member x.Measure msg =
```
haf-raysearchlabs: Remove this one too

src/Logary/Logary.fsproj
```diff
@@ -30,27 +30,37 @@
     <DocumentationFile>bin\Release\Logary.xml</DocumentationFile>
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
-  <Import Project="$(MSBuildExtensionsPath32)\..\Microsoft SDKs\F#\3.1\Framework\v4.0\Microsoft.FSharp.Targets" />
+  <Import Project="$(MSBuildExtensionsPath32)\..\Microsoft SDKs\F#\4.0\Framework\v4.0\Microsoft.FSharp.Targets" />
```
haf-raysearchlabs: Should we upgrade this if we're compiling with 3.1?

src/Logary/Logger.fs
```diff
@@ -1,94 +1,82 @@
...
-      LogLine.setPath loggerName line
-    else line
+  open Logary.DataModel
 
   /////////////////////
   // Logging methods //
   /////////////////////
 
+  let private setContext (logger : Logger) (msg : Message) =
+    Message.Context.serviceSet logger.Name msg
```
haf-raysearchlabs: Not sure about the name `serviceSet`.

src/Logary/Metric.fs
```diff
@@ -25,28 +26,34 @@ type MetricType =
...
-  /// The custom probe shall release any resources associated with the given
-  /// state and return ok.
-  | Shutdown of Acks ReplyChannel
   /// The Reset shall reset the state of the probe to its initial state.
   | Reset
+  /// The custom probe shall release any resources associated with the given
+  /// state and return ok.
+  | Shutdown of IVar<Acks>
+
+type MetricInstance = {
```
haf-raysearchlabs: Coding style:

``` fsharp
type MI =
  { requestCh : ..
    updateCh : ..
     .. }
```

src/Logary/Registry.fs
```diff
@@ -144,51 +134,47 @@ module Advanced =
   /// how that went. E.g. if there's a target is has a backlog, it might not be able to
   /// flush all its pending entries within the allotted timeout (200 ms at the time of writing)
   /// but will then instead return Nack/failed RPC.
-  let flushPending dur (registry : IActor) =
-    if registry.Status <> ActorStatus.Running then async { return Nack "registry not running" }
-    /// Timeout=Infinite below is for THIS call, internally registry has a timeout per target.
-    else registry |> Actor.reqReply (fun c -> FlushPending(dur, c)) Infinite
+  let flushPending dur (registry : RegistryInstance) = job {
```
haf-raysearchlabs: 2 spaces indentation

src/Logary/Registry.fs
```diff
@@ -144,51 +134,47 @@ module Advanced =
...
+        level   = targets |> List.map (fun (_, _, level) -> level) |> List.min
         ilogger = ilogger }
       :> Logger
 
     let wasSuccessful = function
-      | SuccessWith(Ack, _)     -> 0
-      | ExperiencedTimeout _    -> 1
-      | SuccessWith(Nack(_), _) -> 1
+      | Success Ack      -> 0
+      | TimedOut         -> 1
```
haf-raysearchlabs: Use Hopac's timeout instead?

src/Logary/Registry.fs
```diff
@@ -201,37 +187,70 @@ module Advanced =
...
       let regPath = "Logary.Registry.registry"
-      let log = LogLine.setPath regPath >> Logger.log lgr
+      let log = Message.Context.serviceSet regPath >> Logger.log lgr
+
+      /// In the shutdown state, the registry doesn't respond to messages, but rather tries to
+      /// flush and shut down all targets, doing internal logging as it goes.
+      let shutdown state (dur : Duration) (ack : IVar<Acks>) = job {
+        Message.info "shutdown metrics polling" |> log
+
+        do! defaultArg (Option.map (Cancellation.cancel) state.pollMetrics) (Job.unit ())
```
haf-raysearchlabs: Unnecessary brackets

src/Logary/Scheduling.fs
```diff
@@ -4,74 +4,87 @@ module Logary.Internals.Scheduling
...
+
+open Hopac
+open Hopac.Infixes
 
 open NodaTime
 
+type Cancellation = private { cancelled: IVar<unit> }
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Cancellation =
```
haf-raysearchlabs: Nice

src/Logary/Scheduling.fs
```diff
@@ -4,74 +4,87 @@ module Logary.Internals.Scheduling
...
     }
+
     loop ()
 
-/// Create a new scheduler actor
+/// Creates a new scheduler job
 let create () =
-  Actor.spawn (Ns.create "scheduler") Impl.loop
+  let ch = Ch.Now.create ()
+  Job.Global.start (Impl.loop ch)
```
haf-raysearchlabs: I think this can be done better with `server` from Hopac, as it has less overhead (according to docs)

src/Logary/Script.fsx
```diff
@@ -0,0 +1 @@
+﻿
```
haf-raysearchlabs: Remove

src/Logary/Supervisor.fs
```diff
@@ -0,0 +1,121 @@
...
+  | Unregister of string
+  /// Used by a supervisor job to report crashes.
+  /// The supervisor sets the IVar to indicate if the job should be restarted.
+  | JobErrored of name: string * exn: exn option * shouldRestart: IVar<bool>
+
+  type Options =
+    { logger: Logger
+      maxRestarts: int
+      restartDelay: TimeSpan }
+    with
```
haf-raysearchlabs: Unindent

src/Logary/Supervisor.fs
```diff
@@ -0,0 +1,121 @@
...
+
+module Supervisor =
+
+  type internal SupervisorState = Map<string, int>
+
+  type SupervisorMessage =
+  /// Register the given job to be supervised.
+  /// The job will be automatically restarted if it crashes.
+  /// A job can only be registered once.
+  | Register of NamedJob<unit>
```
haf-raysearchlabs: Indent

src/Logary/Internals.fs
paavohuhtala: I wasn't aware that Hopac implements timeouts (other than the `timeOut` Alt, which I used for implementing `Job.withTimeout`). Could you point me in the right direction?

src/Logary/Logger.fs
paavohuhtala: It's a remnant from 3.X; this function was created fairly early in the object model conversion process to make the conversion easier. I'll remove it.

paavohuhtala: Now that I've thought about it more, it's completely wrong. It should set the `name` instead of the `service`, and not overwite it if it's already set.

src/Logary/Internals.fs
haf: I suppose there's no harm in this; I'm suggesting not making it explicit like this, but instead returning an Alt from all operations that can return this DU case and letting the caller worry about whether to time the operation out or not; and having the other side implement selective synchronous operations on its side, effectively making the timeout operation a way that two concurrent processes/jobs/alts can synchronise around something.

haf:
I think it's time to build a prerelease and start testing all of this works with some of our apps :). :100: :+1: so far. I'm having some trouble building, perhaps I can send you a PR to your PR?

paavohuhtala:
Sounds good. I'll merge your PR now and look into more of these during the weekend.

paavohuhtala:
After merging paavohuhtala/logary/pull/1, the core library no longer builds due to the compiler version downgrade. I fixed all of the errors in my code (by replacing a few instances of Map with Map.ofSeq), but it seems that FsMessageTemplates depends on F# 4.0's closure-captured mutables.

haf:
Then I think we should comment out that functionality, so we may replace it with a real parser implementation. What do you say @adamchester ? This is an issue if we'd merge with Suave as well.

haf:
@paavohuhtala By the way; what environment are you in? I can build just fine on Mono 4.2.1, F# 4.0.1.0.

adamchester:
@haf when you say "real parser implementation" what do mean? If I understand correctly what you need is only the parser. I'd like to make this easier to reuse the fast-but-ugly parser (which should rarely change - if ever), but I'd  also like to work on the FParsec version - even using the same tests. Could you clarify some of this please?

adamchester:
Also, if you like, an easy immediate fix is to just switch those few lines to ref cells. I tried to do this just now and submit a PR, but I'm having trouble getting visual studio working right now (thanks VS 2015 update 1!).

It looks like I did the same thing to the parser here: https://github.com/adamchester/MessageTemplates/pull/17/files#diff-c4ca4e3186d4ed811d2469b23c6b5ebaL315

paavohuhtala:
@haf I upgraded Mono, and now it builds fine using Rake/Xbuild, but (obviously) it didn't affect VS's situation. Not building is the right behaviour, though - the .fsproj specifies that it's an F# 3.1 project, so it should be built with the F# 3.1 compiler.

haf:
@adamchester I mean something I can read and understand from reading it... We'd like to add formatting, and I'm also looking at what amount of code a file-include type of API for libraries should look like.

I think we should absolutely add formatting capabilities from you too, for the console target for example!

We should add in your tests and take the latest implementation of your parser. /cc @paavohuhtala

haf:
@paavohuhtala I'm moving everything in `Logary.DataModel` -> `Logary` to keep the consuming API simple.

haf:
@paavohuhtala The reason I wrote "scary" on isRunning is that it always makes you use the API in a manner susceptible to a race condition. In fact, it's impossible to avoid race conditions with that API. Instead, we'd have to use the `timeOut : Alt<unit>` for communicating with things that may hang. The reason we performed the check with actors was that there was no other way of doing selecting communication with them.

haf:
@paavohuhtala Pushed a whole lot of changes to the PR on your repo

paavohuhtala:
@haf 

I knew `isRunning` had problems, but I didn't realize it was that major. I'll merge your changes in a bit, thanks a lot. :smile:

haf:
It should be noted that the current state of my changes doesn't fully compile, but I've cleared away a lot of the previous comments in this thread and I think most changes are for the better, including more compiling and working targets.

The most interesting piece to fix is the flush call; because it returns an Alt, a caller can choose to timeout and do something else; and this will also notify the Registry, which can then skip sending anything on the reply channel. This is what the flush message should look like:

``` fsharp
type RegistryMessage =
  | ...
  | FlushPending of ackCh:Ch<unit> * nack:Promise<unit>
  | ...
```

And similarly so for Target's messages. Shutdown could be done in a similar manner, but it's not very useful because it's not a frequent use-case.

Registry should be changed to:

```
| FlushPending(ackCh, nack) ->
```

By the way, I released the part payment just now. Sorry for the delay, I had a computer crash and had to do a bit of resetting...

paavohuhtala:
Should we use a non-configurable 200ms timeout for `Target.flush` in the registry loop? (ignore the fact the sample still uses `Job.withTimeout`)

``` fsharp
let! allFlushed =
  targets
  |> Seq.pjmap (Target.flush >> Job.withTimeout (Timeout dur)) <-- problem
```

haf:
Yes, sure, that's a good trade-off.

src/Logary/Configuration_Config.fs
```diff
@@ -145,35 +142,45 @@ let runLogary conf =
...
-      member x.Shutdown fDur sDur = shutdown' fDur sDur inst |> run
-      member x.Dispose ()         = shutdown inst |> Async.Ignore |> run
+      member x.runtimeInfo =
+        inst.runtimeInfo
+
+      member x.getLogger name =
+        name |> getLogger inst.registry |> run
+
+      member x.flushPending dur =
+        Advanced.flushPending dur inst.registry |> run
```
paavohuhtala: In the revised interface this function (and probably `shutdown` too) should return an `Alt<unit>`, instead `run`ning the job. Would that mean it should no longer start the job, and leave that for the caller?

haf: On an interface level yes. On a "Logger Extensions" level, there would be another method that also ran it, which will be more common to use.

Before spending any large amounts of time on this I recommend you to fix the cyclic message buffers (overwrite oldest Message), because then this becomes mostly a non-issue.

haf-raysearchlabs:
Relevant: https://github.com/Hopac/Hopac/issues/87

haf:
Found a bit of funky code:

``` fsharp
  let isCancelled cancellation = job {
    return! (IVar.read cancellation.cancelled ^->. true) <|> (Alt.always false)
  }
```

This will create a job that blocks until the cancellation has been cancelled, because IVar.read downcasts to Alt<_> which in turn only yields when the IVar is set.

paavohuhtala:
@haf

I did a small repro in REPL, and it works as expected. If `IVar.read` isn't available immediately, `Alt.always false` is picked instead.

```fsharp
> let ivar : IVar<unit> = IVar.Now.create ();
	val ivar : IVar<unit>
> Job.Global.run (IVar.read ivar ^->. true <|> Alt.always false);;
	val it : bool = false
> Job.Global.run (IVar.fill ivar ());;
	val it : unit = ()
> Job.Global.run (IVar.read ivar ^->. true <|> Alt.always false);;
	val it : bool = true
```

haf:
Ah, excellent

src/Logary/Configuration_Config.fs
```diff
@@ -140,40 +137,48 @@ let validate ({ targets  = targets
...
-      member x.Shutdown fDur sDur = shutdown' fDur sDur inst |> run
-      member x.Dispose ()         = shutdown inst |> Async.Ignore |> run
+      member x.runtimeInfo =
+        inst.runtimeInfo
+
+      member x.getLogger name =
+        name |> getLogger inst.registry |> run
+
+      member x.flushPending dur =
+        Advanced.flushPending inst.registry <|> timeOut (dur.ToTimeSpan ())
```
polytypic: Note that is usually preferable to have `Alt` operations without baked in timeouts.  In other words, instead of `flushPending: Duration -> Alt<unit>` you could have `flushPending: Alt<unit>`.  The user of `flushPending` would still be able to combine with a `timeOut` or run the operation without a timeout.

src/Logary/Metric.fs
```diff
@@ -70,65 +75,61 @@ with
...
 /// The GetValue implementation shall retrieve the value of one or more data
 /// points from the probe.
-let getValue (dps : DP list) =
-  Actor.reqReply
-    (fun chan -> GetValue(dps, chan))
-    Infinite
+let getValue (dps : PointName list) mi =
+  let resultCh = Ch ()
+  mi.requestCh *<+ GetValue (dps, resultCh) >>=.
+  resultCh
```
polytypic: The above has async client request paired with sync server reply, but without nack.  This is an unusual pattern for Hopac.

Typically one wants either

* async client request with nack (`reqCh *<+ (nack, repCh)`) and sync server reply (`repCh *<- result <|> nack`)

or

* sync client request (`reqCh *<- repIv`) and async server reply (`repIv *<= result`).

The first one has commit point on server reply and the second one has commit point on client request.  Both patterns are alternatives and can then be combined using `Alt` combinators.

src/Logary/Registry.fs
```diff
@@ -144,51 +137,55 @@ module Advanced =
...
   /// but will then instead return Nack/failed RPC.
-  let flushPending dur (registry : IActor) =
-    if registry.Status <> ActorStatus.Running then async { return Nack "registry not running" }
-    /// Timeout=Infinite below is for THIS call, internally registry has a timeout per target.
-    else registry |> Actor.reqReply (fun c -> FlushPending(dur, c)) Infinite
+  let flushPending (registry : RegistryInstance) : Alt<unit> = Alt.withNackJob <| fun nack -> job {
+    let! flushCh = Ch.create ()
+    do! registry.reqCh *<+ (FlushPending (flushCh, nack))
+    return flushCh
+  }
```
polytypic: The above looks like a proper selective client side operation.  Note that the [*<+->-](http://hopac.github.io/Hopac/Hopac.html#def:val Hopac.Infixes.*<+->-) operator directly addresses this case.  You could just write

```fsharp
let flushPending (registry : RegistryInstance) : Alt<unit> =
  registry.reqCh *<+->- fun flushCh nack -> FlushPending (flushCh, nack)
```

You can think of the `*<+->-` operator (and other similar operators) as performing a call to an concurrent server.

src/Logary/Scheduling.fs
```diff
@@ -4,74 +4,90 @@ module Logary.Internals.Scheduling
...
+type Cancellation = private { cancelled: IVar<unit> }
+
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module Cancellation =
+  let create () =
+    { cancelled = IVar.Now.create () }
+
+  // https://github.com/logary/logary/pull/92#issuecomment-164139379
+  let isCancelled cancellation = job {
+    return! (cancellation.cancelled ^->. true) <|> (Alt.always false)
```
polytypic: This, using a promise (or ivar) as a cancellation token like thing, is actually one of the valid cases where one could use [isFulfilled](http://hopac.github.io/Hopac/Hopac.html#def:val Hopac.Promise.Now.isFulfilled).

src/Logary/Scheduling.fs
```diff
@@ -4,74 +4,90 @@ module Logary.Internals.Scheduling
...
 open System.Threading
-  
-open FSharp.Actor
+
+open Hopac
+open Hopac.Infixes
 
 open NodaTime
 
+type Cancellation = private { cancelled: IVar<unit> }
```
polytypic: Yes, this will work, but I would suggest a slightly different design.  Specifically, only expose cancellation as a `Promise<unit>`.  This allow plugging in a nack (from withNack) as the cancellation token.

src/Logary/Scheduling.fs
```diff
@@ -4,74 +4,90 @@ module Logary.Internals.Scheduling
...
+  let create () =
+    { cancelled = IVar.Now.create () }
+
+  // https://github.com/logary/logary/pull/92#issuecomment-164139379
+  let isCancelled cancellation = job {
+    return! (cancellation.cancelled ^->. true) <|> (Alt.always false)
+  }
+
+  let cancel cancellation = job {
+    do! IVar.fill cancellation.cancelled ()
```
polytypic: For the particular case of cancellation I would recommend using `IVar.tryFill`.

Note also the comment above about using `Promise<unit>` rather than `IVar<unit>` as the type.  You could then use a design similar to what .Net does by having a cancellation token "source".  The source knows the underlying `IVar`, while the the token itself only exposes it as a `Promise`.

See also [here](https://github.com/Hopac/Hopac/issues/33#issuecomment-67754873).

src/Logary/Internals.fs
```diff
@@ -1,27 +1,98 @@
...
-    |> fun str -> Actor.Options.Create str
+    [ LogaryRootNs; subcomponent ]
+    |> NamedJob.create
 
 module internal Seq =
+
   let all f s = Seq.fold (fun acc t -> acc && f t) true s
   let any f s = Seq.fold (fun acc t -> acc || f t) false s
-  let pmap (f : _ -> Async<_>) s = s |> Seq.map f |> Async.Parallel
+  let pjmap f s = s |> Seq.map f |> Job.conCollect
```
polytypic: Note that [Seq.Con.mapJob](http://hopac.github.io/Hopac/Hopac.html#def:val Hopac.Extensions.Seq.Con.mapJob) does the same thing.

src/Logary/Registry.fs
polytypic: While the pattern using withNack above is correct, on deeper look, it would seem that for this operation nack is unnecessary.  It would seem that this operation (flushing) is essentially an idempotent operation.  Therefore you should be able to do with just `prepare`.  Something like this:

```fsharp
let flushPending (registry : RegistryInstance) : Alt<unit> =
  Alt.prepareJob <| fun () ->
  let flushIv = IVar ()
  registry.reqCh *<+ FlushPending flushIv >>-
  flushIv
```

src/Logary/Registry.fs
```diff
@@ -201,51 +198,94 @@ module Advanced =
...
+
+        if List.isEmpty failed then
+          do! Message.info "shutdown Ack" |> log
+        else
+          let msg = sprintf "failed target shutdowns:%s%s" nl (toTextList failed)
+          do! Message.errorf "shutdown Nack%s%s" nl msg
+              |> Message.field "failed" (Value.fromObject failed)
+              |> log
+
+        do! Ch.give ackCh () <|> nack
```
polytypic: Note that here the semantics don't quite work out, because even if the nack is set (client no longer waits) the shutdown isn't cancelled.  Use nack when the server can (and should) cancel the operation when client stops waiting.

One way to treat the shutdown is as an idempotent (and non cancellable) operation.  In that case you would use `prepare` to async send (`*<+`) an `IVar` to the server that the server then fills when shutdown is complete.

More complex protocols are also possible.  In this case it would likely be possible to make a kind of two-phase shutdown protocol.  Things that need to be shutdown could be in three states: "active" (accepting all messages and acting upon them), "shutting down" (not accepting messages, but still having resources intact) and "disposed" (not accepting messages/responding with error and having resources released).  Upon receiving an asynchronously sent shutdown request, containing a nack, things would go into "shutting down" state by flushing resources and making it trivial to dispose everything.  After everything is "ready to flip the dispose switch", the client would be acknowledged synchronously and everything would be disposed or, in case client has aborted (nack is set and client does not take ack message), everything would be transferred back to active case (meaning that shutdown is cancelled).

haf:
Closing in favour of #90

PR #110, opened 2016-04-27
Feature/middleware

haf:
This implements middlware for logary and brings the C# API back in line with v4. It also adds the `time` function to the `Logging` module.


src/Logary.CSharp/LoggerExtensions.cs
```diff
@@ -106,203 +108,41 @@ public static T Time<T>(this Logger logger, Func<T> f, LogLevel level = null)
...
-            level = level ?? LogLevel.Debug;
-            a();
+            Time (logger, () => {
+                a ();
+                return 0;
+            }, level);
         }
 
         /// <summary>
         /// TBD
```
neoeinstein: Update the XML doc now that the implementation is provided.

src/Logary/Configuration_Config.fs
```diff
@@ -104,6 +105,14 @@ let withMetrics ms conf =
   |> Seq.map Metric.validate
   |> Seq.fold (fun s t -> s |> withMetric t) conf
 
+[<CompiledName "WithMiddleware">]
+let withMiddleware middleware conf =
+  { conf with LogaryConf.middleware = middleware :: conf.middleware }
```
neoeinstein: Does/will ordering of middlewares matter? This looks to be consistent with your concatenation of multiple middlewares below, but we will want to be explicit about how the ordering affects the application.

haf: Middleware registered first will run first. So

``` fsharp
|> withMiddleware messageId
|> withMiddleware principalDetails
```

will first run messageId, then `principalDetails`. So if `messageId` is the same middleware as we have in logary-js, then this order would be in error.

futurechan:
👍

PR #114, opened 2016-04-29
Feature/middleware

whyer:
Now logary actually uses the configured middleware!

### DIFF
--- a/src/Logary/LogManager.fs
+++ b/src/Logary/LogManager.fs
@@ -11,7 +11,7 @@ open Logary.Middleware
 /// running targets.
 type RegistryMessage =
   /// Get a logger for the given point name (the path of the logger)
-  | GetLogger of logger:PointName * middleware:(unit -> Message -> Message) * replCh:IVar<Logger>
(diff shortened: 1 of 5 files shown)


haf:
Nice!

haf:
I was a bit too early here. You tests cause the test runner to hang.

haf:
I've reverted the merge, please resend it

whyer:
I am very sorry but I can not reproduce any hangs on my Windows machines (at home or work). Maybe some differences with mono-implementation on OSX?

Can you point out some code that may be wrong?

haf:
If you resend the PR I've added travis integration.

The only change of the code is the code that sends GetLogger to Registry. I think it may be connected to when the 'run' in the testCase is being called with respect to when the GetLogger message is received by the Registry.

The solution then would be to replace 'run' and have a `job { ... }` around it all which is run at the end of the testCase, returning the values to be tested.

This hasn't been required for any other test though, so if you PR again then we can see what travis says.

PR #118, opened 2016-05-10
Serilog adapter

adamchester:
This does not yet work. It does not pass Serilog event properties into the Logary format.

It is based on the log4net adapter.

At this stage, it just an experiment.


haf:
Sweet. Sorry about the state of CI, it's all pretty new. @whyer is the guy working on this this week. This is also a good opportunity to tell Andreas that I'll be around accepting PRs if that's needed. ;)

haf:
Trying to rebuild this PR against current master.

adamchester:
Rebased against master

haf:
Ready to merge?

haf:
Hmm just noticed there were no tests... Having such is probably a good idea before merging :)

haf:
Will need rebasing.

adamchester:
OK rebased.

It definitely needs a bit of testing before merging I think, because I'm not sure about how it passes the structured values from Serilog -> Logary right now.

I need to find a way to test it out properly. Will make another attempt soon.

adamchester:
@haf this makes me sad :(

Failed on Mono 4.2.3: https://travis-ci.org/logary/logary/jobs/150207968#L1279
Passed on Mono 4.4.1: https://travis-ci.org/logary/logary/jobs/150207969#L1273

adamchester:
OK I think this is ready to be tested out in the wild.

thoughts?

adamchester:
@haf I have a feeling i'm not flushing the logary pipeline correctly. any ideas?

haf:
Yes, you're right that you're not flushing the pipeline correctly. It's because you're using `logSimple` in the adapter which puts the work on a global Hopac queue thereby making it asynchronous with respect to the caller and targets in Logary. If Serilog doesn't have the concept of acks (successful persistance of a log message) then the next step is to log something to your Logary target, also with `start` which guarantees order, and then poll the output until your own 'in the unit test' message is visible on the output; then you can assert on the log message sent through Serilog.

adamchester:
This was an old experiment. I never had a strong enough reason to see this through to completion. If anyone has a need for this some day, I'd be willing to help out.

PR #119, opened 2016-05-10
Allow named fields in `eventFormat` templates 

adamchester:
The goal is to allow [`eventFormat`](https://github.com/logary/logary/blob/f5c3e31fcf96678a7b02db8f59c365893cfcbc6d/src/Logary/DataModel.fs#L1178) to accept log messages like:

```fsharp
testCase "Formatting.templateFromFormat, named fields" <| fun _ ->
    let format = "This {gramaticalStructure} contains {wordCount} words."
    let args : obj[] = [|"sentence"; 4|]
    (because "fields are matched left-to-right in message template" <| fun () ->
        Message.templateFromFormat format args)
    |> should equal ("This {gramaticalStructure} contains {wordCount} words.",
                        [ "gramaticalStructure", Field (String "sentence", None)
                        "wordCount", Field (Int64 4L, None) ])
    |> thatsIt
```

### DIFF
--- a/src/Logary/DataModel.fs
+++ b/src/Logary/DataModel.fs
@@ -1182,9 +1182,26 @@ module Message =
   /// a message template and a set of fields.
   let internal templateFromFormat (format : string) (args : obj[]) =
     let parsedTemplate = Parser.parse format
-    let scalarNamesAndValues = captureNamesAndValuesAsScalars parsedTemplate args
+    let isPositional = parsedTemplate.Positionals <> Unchecked.defaultof<PropertyToken[]>
+    let positionalFixedTemplate =
+      if isPositional then
+        // if the properties are all positional {0}, {1}..{N} then convert them {arg0}, {arg1}..{argN}
+        let collectedProps = ResizeArray<PropertyToken>(parsedTemplate.Positionals.Length)
+        let prependArgName (pt : PropertyToken) =
+          let p = PropertyToken("arg" + pt.Name, pt.Pos, pt.Destr, pt.Align, pt.Format)
+          collectedProps.Add(p)
+          p
+
+        let tokenOrPrependArgNameProp t = match t with Text _ as tt -> tt | Prop (start, pt) -> Prop(start, prependArgName pt)
+        let tokensWithArgNameInProps = parsedTemplate.Tokens |> Seq.map tokenOrPrependArgNameProp |> Array.ofSeq
+        let newFormatString = String.Join("", tokensWithArgNameInProps |> Array.map string)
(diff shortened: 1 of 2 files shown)


haf:
Cool, and I'm sure @mavnn has a pointer or two about string extrapolation at compile time =)

src/Logary/DataModel.fs
```diff
@@ -1158,19 +1158,33 @@ module Message =
...
+  open Logary.Utils.FsMessageTemplates
+
+  /// A destructuring strategy for FsMessageTemplates which simply treats 
+  /// everything as a 'Scalar' object which can later be handled by Logary
+  /// as `Field (Value.ofObject o, None)`
+  let internal destructureEverythingAsScalarObject : Destructurer =
+    fun request -> TemplatePropertyValue.ScalarValue request.Value
+  
+  let internal captureNamesAndValuesAsScalars (t: Template) (args: obj[]) = 
+    let selfLogByIgnoring : SelfLogger = fun (t, v) -> ()
```
adamchester: @haf I was wondering about what to do with `SelfLogger` - as you can see on this line, it is currently ignored.

Take the following scenario as an example:
```fsharp
    Message.eventFormat (level, "Hello, {one} {two}", [| 1; 2; 3 |]) // oops - 2 fields <> 3 params
    |> Logger.log myLogger
    |> start
```

I've specified that there will be two parameters (`{one}` and `{two}`), however I actually provided three (`[| 1; 2; 3 |]`). Somebody should be warned about this mistake, but I'm not sure to go about that here. It looks like Logary has an `internal target` but I don't seem to have access to it here. Any ideas?

logarynamedtemplatefields-warning

haf:
How about making it a `Field ('missing template variable for position', Int64 3L`

adamchester:
@haf the idea of including a field is good.

src/Logary/DataModel.fs
adamchester: I am considering ignoring this issue in the short term, to get this moving along. What do you think @haf ?

haf:
Restarting build to use master.

haf:
Now it shows a valid failure related to the formatting. =p

adamchester:
Ok, I added back the old behaviour so positionals have the `argN` but I'm thinking its a bit too ugly :cry: 

Maybe I'll have another shot at cleaning it up, if you think it's worthwhile

adamchester:
@haf this is how I think it should work. Let me know if you disagree.

haf:
I think this is looking good :)

PR #121, opened 2016-05-11
Feature/strong naming

whyer:
Now using new hopac-strongname nuget package that I published.


haf:
Can you build a local package in the mean while?

PR #125, opened 2016-06-07
[WIP] updated Logary.Target.Zipkin

Horusiath:
WORK IN PROGRESS (do not merge)

This is bundled and refactored target for Zipkin using [Zipkin.Tracer](https://github.com/bazingatechnologies/Zipkin.Tracer) library (only locally right now, it will be referenced once the library will be released).

Please review and comment.


src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
```diff
@@ -0,0 +1,141 @@
...
+        /// Default port: 9411
+        /// Default host: localhost
+        static member Create(?baseUri, ?fac, ?traceReq, ?formatter) = 
+            { baseUri = defaultArg baseUri (Uri "http://localhost:9411")
+              collectorFactory = defaultArg fac (fun uri -> upcast HttpCollector uri) }
+        
+        static member Default = ZipkinConf.Create()
+    
+    type private ZipkinState = 
+        { activeSpans : Map<uint64, Span>
```
Horusiath: This is probably wrong. I need some advice on how we can carry span for current RPC request.

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
```diff
@@ -0,0 +1,141 @@
...
+          collector : ISpanCollector }
+    
+    let private makeSpan (message: Message): Span = 
+        let (Int64 traceId) = message.context.["traceId"]
```
Horusiath: To be created span needs IDs of the current trace (serie of rpc requests started by some user action) and current span (which determines part of the overall distributed RPC request executed by current service). Additionally service localization (ip:port) is required.

It looks like the only place for that is message context.

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
```diff
@@ -0,0 +1,141 @@
...
+                             | _ -> ()
+                             (ack *<= () :> Job<_>)
+                         RingBuffer.take requests ^=> function 
+                         | Log(logMsg, ack) -> 
+                             job { 
+                                 let nextState = annotateSpan state logMsg
+                                 do! ack *<= ()
+                                 return! loop nextState
+                             }
+                         | Flush(ack, nack) ->
```
Horusiath: Once RPC request ends, span needs to be released and send with all of the annotations back to zipkin. Is it a good idea to do that for `Flush` message?

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
haf: It's how I would do it, but I would use `Scheduler` to schedule removal of the span after a timeout. You could also use the Cache classes which support timed removal.

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
haf: Yes that's right.

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
```diff
@@ -0,0 +1,141 @@
...
+                                 return! loop nextState
+                             }
+                         | Flush(ack, nack) -> 
+                             job { 
+                                 let spans =
+                                    state.activeSpans
+                                    |> Map.toSeq
+                                    |> Seq.map snd
+                                    |> Seq.toArray
+                                 do! collector.CollectAsync spans |> Async.AwaitTask
```
haf: You probably don't need to AwaitTask here, because Hopac supports binding to `Task<>`.

src/targets/Logary.Targets.Zipkin/Targets_Zipkin.fs
haf: No that's not what it's designed to do. It's better to flush when you receive a message about EndRequest matching an active span.

PR #126, opened 2016-06-09
Feature/c sharp fixes

whyer:
Some fixes for the C# extensions methods.
Also adding mspec runner to rakefile. Interesting to see if it works with mono...


haf:
@whyer Thanks!

PR #127, opened 2016-06-27
logstash target implement pubsub pattern

ademar:


haf:
Thanks, this is great. Looking forward to merging. However, shouldn't it remain `connect` since configuration for logstash is configuration managed and written once. Not re-written for every extra node that may want to send data to logstash, which would be the case if the `SUB` socket on logstash would need to work in [`client`](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-zeromq.html#plugins-inputs-zeromq-mode) mode and we'd be forced to have a new such zeromq input for every thing we'd like to collect logs from?

For reference:

 - https://github.com/zeromq/netmq/issues/94

ademar:
I do not follow exactly what you mean here 

>  shouldn't it remain connect since configuration for logstash 

In this config logary is a publisher and logstash the subscriber so we need to `bind`

```text
input {
   zeromq {
    topology => "pubsub"
    mode => "client"
    address => ["tcp://127.0.0.1:2120"]
  }
}
output {
  elasticsearch { hosts => [ "localhost:9200"] }
  stdout { codec => rubydebug }
}

```

I understand that there are different configurations and you may want the target to support all of them. but I do not have the time to do so.

With this changes at least you get one "pubsub" working configuration whilst before it did not worked at all .. YMMV

haf:
What I mean is; if you configure logstash as client it would need to know where from to consume, which will potentially be many different addresses... right?

ademar:
OK, got it working in "pushpull" mode. I'll make it so that it is configurable since I already started  :)

ademar:
@haf what about now ? "pushpull" is default and works out of the box with a config like :

```text
input {
  zeromq {
    topology => "pushpull"
  }
}
```

haf:
Thanks @ademar :) We'll merge during the day.

mavnn:
Now released (apart from Rutta)

PR #131, opened 2016-06-28
Added logging of Event to InfluxDB

whyer:
- logging Template as "Event" tag
- logging Fields as InfluxDB fields

Useful for annotations in InfluxDB metrics!


haf:
Thanks @whyer !

PR #132, opened 2016-06-29
[infra] Build mono v4.4.1

haf:
Broken mono version.


haf:
Ref https://github.com/fsharp/fsharp/issues/583

PR #134, opened 2016-07-07
[DataModel] Fixing NullReference exception in Value.ofObject

whyer:
Also added some FsCheck property tests for the same function


haf:
Thanks @whyer !

PR #145, opened 2016-08-05
Attempt to fix build/test on windows

adamchester:
by adding `app.config` with the `FSharp.Core` binding redirect


haf:
Almost there now...

PR #149, opened 2016-08-13
Grammar/typo tweaks in README

bartelink:
The readme is pretty clear as it is, so not very many interesting changes here.

I probably wouldn't move it up, but it may be worth calling out in the opening summary that there is a comparison to NLog/log4net section that may be helpful for those not even familiar with semantic logging etc.

Some notes re items I didn't address:

In the

   `let msg = Message.gauge pn (Float value)`

uses n undefined `pn` (and I dont understand enough to 'just' fix it :)

The `Kafunk` ref is nice, but may be meaningless to the average reader so suggest using `Libryy`
(Aah, looking at it again, your intent is clear as the rake equivalent doesn't refer to it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,10 @@ Logary is a high-performance, semantic logging, health and metrics library for
  - Full support for Semantic Logging
  - F# idiomatic code
  - Using C#? Then `Logary.CSharp` is for you!
- - Looking for an F# alternative to LibLog?
-   [`Logary.Facade`](#using-logary-in-a-library) is for you.
+ - Looking for an F# alternative to [`LibLog`](https://github.com/damianh/LibLog)?
+   Jump to [`Logary.Facade`](#using-logary-in-a-library).
  - Never throws exceptions
- - Low latency logging – evaluate your Message only if level is met
+ - Low overhead logging – evaluate your Message only if a level is switched on
  - Supports [Hierarchical logging](#rule--hierarchical-logging)
  - Add metrics to your service/app!
  - A wide range of adapters and targets to choose from!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -409,24 +410,24 @@ Message.event "Hi {user}!"
...
 Message Templates are a superset of standard .NET format strings, so any format
 string acceptable to string.Format() will also be correctly processed by
 FsMessageTemplates.
 
- * Property names are written between { and } brackets
- * Brackets can be escaped by doubling them, e.g. {{ will be rendered as {
- * Formats that use numeric property names, like {0} and {1} exclusively, will
-   be matched with the Format method's parameters by treating the property names
-   as indexes; this is identical to string.Format()'s behaviour
+ * Property names are written between `{` and `}` braces
```
haf: Interesting. Made me [read through this](https://english.stackexchange.com/questions/3379/bracket-vs-brace).

README.md
```diff
@@ -266,32 +267,32 @@ type Rule =
     messageFilter : MessageFilter }
 ```
 
-And you can find the configuration on the module with the same name. The
-`Rule.empty` value is one where you accept any logs from anything.
+You can find the configuration in the module with the same name. The
+`Rule.empty` value is a null one that accepts all logs from anything.
```
haf: I always have a bit of a problem with all vs any. Wouldn't `any logs from anything` subsume `all logs from anything`, in a sense?

README.md
```diff
@@ -438,8 +439,8 @@ metric's computation function at a regular interval, and also has provisions
 for sending your metric other metrics, so that you can chain metrics
 together.
 
-The `ticker` is where you should return Messages (Gauge or Derived values) and
-keep track of how 'far long' you've computed, as to avoid returning the same
+The `ticker` is where you return Messages (Gauge or Derived values) to
+keep track of how 'far along' you've reached, in order to avoid returning the same
```
haf: Another [interesting change](https://ell.stackexchange.com/questions/11517/to-versus-in-order-to), thanks.

haf:
I love your fixes. Going through them all whilst trying to teach myself their justifications, now.

README.md
bartelink: Good find. Don't agree with the non-US English kvetching comments on it.

README.md
bartelink: Perhaps how 'far along' should be progress (I'd have gone for that if I actually understood from that text what it does).

Perhaps "to avoid returning the same message multiple times" should be "to signify that the newest instances of a message should be considered to supersede and/or logically be coalesced with previous occurrences of that message"

As I say, right now it doesnt convey to me the nature of the mechanism

PR #152, opened 2016-08-24
Simplifiy `setField`

neoeinstein:
Removes another instance of a `Map.remove k |> Map.add k v` chain. Additionally simplifies out `put` since `Map.add` handles that case appropriately.


haf:
Thanks @neoeinstein

PR #155, opened 2016-08-31
Facade literate console target

adamchester:
(NOTE: Work in progress - seeking feedback now).

When the facade is used by projects like Suave, where it's reasonable to expect a developer to want first-class console logging output (at least during design/development time), it's important to have great human-readable console output.

In this PR, I propose a `LiterateConsoleTarget` be included. When using the `Literate` console target, the output should be optimised for readability (using colours where appropriate), and preferably render with dates, currency, and other formats based on the `Current` culture (if logging to a text stream, one would obviously prefer things like UTC and possibly `Invariant` depending on the circumstances). This requires the `{property:format}` syntax (like `String.Format` or `printf` does) to be supported when parsing message templates. Additionally, it's reasonably to expect users don't want to pay a big price for using template-based structured logging, so it needs to remain reasonably fast.

I believe this feature can help drive adoption of the facade in the F# community, providing a great reason to use embedded (structured) fields in logging statements, 

Here is a sample I used with Suave output, to show how it currently looks:
image

* This is heavily based on my favourite console output theme, the amazing [Serilog Literate Console](https://github.com/serilog/serilog-sinks-literate/).


src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
 
+/// The output tokens, which can be potentially coloured.
+type LiterateToken =
+  | Text | Subtext
+  | Punctuation
+  | LevelVerbose | LevelDebug | LevelInfo | LevelWarning | LevelError | LevelFatal
+  | KeywordSymbol | NumericSymbol | StringSymbol | OtherSymbol | NameSymbol
+
+type LiterateContext =
+  { FormatProvider: IFormatProvider
```
haf: `pascalCase` please

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+              | LevelWarning -> ConsoleColor.Yellow
+              | LevelError -> ConsoleColor.Red
+              | LevelFatal -> ConsoleColor.Red
+              | KeywordSymbol -> ConsoleColor.Blue
+              | NumericSymbol -> ConsoleColor.Magenta
+              | StringSymbol -> ConsoleColor.Cyan
+              | OtherSymbol -> ConsoleColor.Green
+              | NameSymbol -> ConsoleColor.Gray
+      PrintTemplateFieldNames = false }
+  static member CreateInvariant() =
```
haf: Same here

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+    LiterateContext.Create(Globalization.CultureInfo.InvariantCulture)
+
+module internal FsMtParser =
+  open System.Text
+
+  type Property(name:string, format: string) =
+      static let emptyInstance = Property("", null)
+      static member Empty = emptyInstance
+      member __.Name = name
```
haf: Use `x` not `__`

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+  module internal Internal =
+
+    let inline isLetterOrDigit c = System.Char.IsLetterOrDigit c
+    let inline isValidInPropName c = c = '_' || System.Char.IsLetterOrDigit c
+    let inline isValidInFormat c = c <> '}' && (c = ' ' || isLetterOrDigit c || System.Char.IsPunctuation c)
+    let inline isValidCharInPropTag c = c = ':' || isValidInPropName c || isValidInFormat c
+
+    [<Struct>]
+    type Range(startIndex:int, endIndex:int) =
+        member inline this.Start = startIndex
```
haf: 2 space indentation

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+module internal FsMtParser =
+  open System.Text
+
+  type Property(name:string, format: string) =
+      static let emptyInstance = Property("", null)
+      static member Empty = emptyInstance
+      member __.Name = name
+      member __.Format = format
+      member internal x.AppendPropertyString(sb: StringBuilder, ?replacementName) =
+          let replacementName = defaultArg replacementName name
```
haf: Spaces

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+
+    [<Struct>]
+    type Range(startIndex:int, endIndex:int) =
+        member inline this.Start = startIndex
+        member inline this.End = endIndex
+        member inline this.Length = (endIndex - startIndex) + 1
+        member inline this.GetSubString (s:string) = s.Substring(startIndex, this.Length)
+        member inline this.IsEmpty = startIndex = -1 && endIndex = -1
+        static member Substring (s:string, startIndex, endIndex) = s.Substring(startIndex, (endIndex - startIndex) + 1)
+        static member Empty = Range(-1, -1)
```
haf: `pascalCase`

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+
+    let inline tryGetFirstCharInRange predicate (s:string) (range:Range) =
+        let rec go i =
+            if i > range.End then -1
+            else if not (predicate s.[i]) then go (i+1) else i
+        go range.Start
+
+    let inline tryGetFirstChar predicate (s:string) first =
```
haf: `(s : string)`

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+            Property.Empty
+        elif (not formatRange.IsEmpty) && (hasAnyInRange (not<<isValidInFormat) template formatRange) then
+            Property.Empty
+        else
+            let format = if formatRange.IsEmpty then null else formatRange.GetSubString template
+            Property(propertyName, format)
+
+    let inline findNextNonPropText (startAt: int) (template: string) (foundText: string->unit) : int =
+        let rec go i =
+            if i >= template.Length then
```
haf: Indentatation

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+                | '{' -> if (i+1) < template.Length && template.[i+1] = '{' then go (i+2) else i
+                | '}' when (i+1) < template.Length && template.[i+1] = '}' -> go (i+2)
+                | _ -> go (i+1)
+        let nextIndex = go startAt
+        if (nextIndex > startAt) then
+            foundText (Range.Substring(template, startAt, nextIndex - 1))
+        nextIndex
+
+    let findPropOrText (start:int) (template:string)
+                        (foundText: string->unit) (foundProp: Property->unit) : int =
```
haf: Spaces

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+        let first = start
+        let nextInvalidCharIndex =
+            match tryGetFirstChar (not << isValidCharInPropTag) template (first+1) with
```
haf: Indentation

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+            | -1 -> template.Length | idx -> idx
+
+        if nextInvalidCharIndex = template.Length || template.[nextInvalidCharIndex] <> '}' then
+            foundText (Range.Substring(template, first, (nextInvalidCharIndex - 1)))
+            nextInvalidCharIndex
+        else
+            let nextIndex = nextInvalidCharIndex + 1
+            let propInsidesRng = Range(first + 1, nextIndex - 2)
+            match tryGetPropInRange template propInsidesRng with
+            | prop when not (obj.ReferenceEquals(prop, Property.Empty)) -> foundProp prop
```
haf: Let's keep to the formatting guidelines, with newlines after `->`.

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+            | _ -> foundText (Range.Substring(template, first, (nextIndex - 1)))
+            nextIndex
+
+  let parseParts (template:string) foundText foundProp =
+      let tlen = template.Length
+      let rec go start =
+          if start >= tlen then ()
```
haf: Indentation

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+            nextInvalidCharIndex
+        else
+            let nextIndex = nextInvalidCharIndex + 1
+            let propInsidesRng = Range(first + 1, nextIndex - 2)
+            match tryGetPropInRange template propInsidesRng with
+            | prop when not (obj.ReferenceEquals(prop, Property.Empty)) -> foundProp prop
+            | _ -> foundText (Range.Substring(template, first, (nextIndex - 1)))
+            nextIndex
+
+  let parseParts (template:string) foundText foundProp =
```
haf: Spaces

src/Logary.Facade/Facade.fs
haf: Optimisation?

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+      let tlen = template.Length
+      let rec go start =
+          if start >= tlen then ()
+          else match Internal.findNextNonPropText start template foundText with
```
haf: Too much indentation

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
-    fieldRegex.Matches template
-    |> Seq.cast<Match>
-    |> Seq.toList
-    |> List.choose (fun m ->
-      let name = m.Value.Trim([| '{'; '}'; ' '|])
-      if fields |> Map.containsKey name then
-        Some (name, fields |> Map.find name, m)
-      else
-        None)
+  let literateFormatValue (context: LiterateContext) (fields: Map<string,obj>) = function
```
haf: Spaces

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+            | :? int16 | :? int32 | :? int64 | :? decimal | :? float -> NumericSymbol
+            | :? string | :? char -> StringSymbol
+            | _ -> OtherSymbol
+          if context.PrintTemplateFieldNames then
+            themedParts.Add ("["+prop.Name+"] ", Subtext)
+          matchedFields.Add prop.Name
+          themedParts.Add (fieldAsText, valueColour)
+
+        | None ->
+          // if the property wasn't provided, just render the missing value
```
haf: Have you added unit tests for all these cases?

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+          // but using 'LevelFatal' so it really stands out.
+          themedParts.Add (prop.ToString(), LevelFatal)
+
+      FsMtParser.parseParts template foundText foundProp
```
haf: I don't understand this code without code-completion - does `FsMtParser.parseParts template foundText foundProp` return a function that you give a set to? Or are you mutating something?

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+      Set.ofSeq matchedFields, List.ofSeq themedParts
+
+    | Gauge (value, units) ->
+      Set.empty, [ sprintf "%i" value, NumericSymbol
+                   sprintf "%s" units, KeywordSymbol ]
+
+  let formatValue (fields: Map<string,obj>) (pv: PointValue) =
+    let matchedFields, themedParts =
+      literateFormatValue (LiterateContext.CreateInvariant()) fields pv
+    matchedFields,  System.String.Join("", themedParts |> List.map fst)
```
haf: `System.String.Concat`?

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+    let inline hasAnyInRange predicate (s:string) (range:Range) =
+        match tryGetFirstChar (predicate) s range.Start with
+        | -1 -> false | i -> i <= range.End
+
+    let inline hasAny predicate (s:string) =
+        hasAnyInRange predicate s (Range(0, s.Length - 1))
+
+    let inline indexOfInRange s range c = tryGetFirstCharInRange ((=) c) s range
+
+    let inline tryGetPropInRange (template:string) (within : Range) : Property =
```
haf: This is a log of code that's pretty opaque. Could you explain using some code comments how it's put together?

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+        if propertyName = "" || (hasAny (not<<isValidInPropName) propertyName) then
+            Property.Empty
+        elif (not formatRange.IsEmpty) && (hasAnyInRange (not<<isValidInFormat) template formatRange) then
+            Property.Empty
+        else
+            let format = if formatRange.IsEmpty then null else formatRange.GetSubString template
+            Property(propertyName, format)
+
+    let inline findNextNonPropText (startAt: int) (template: string) (foundText: string->unit) : int =
+        let rec go i =
```
haf: Needs docs

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+        let propertyName = nameRange.GetSubString template
+        if propertyName = "" || (hasAny (not<<isValidInPropName) propertyName) then
+            Property.Empty
+        elif (not formatRange.IsEmpty) && (hasAnyInRange (not<<isValidInFormat) template formatRange) then
+            Property.Empty
+        else
+            let format = if formatRange.IsEmpty then null else formatRange.GetSubString template
+            Property(propertyName, format)
+
+    let inline findNextNonPropText (startAt: int) (template: string) (foundText: string->unit) : int =
```
haf: Needs docs

src/Logary.Facade/Facade.fs
```diff
@@ -226,27 +226,194 @@ type LoggingConfig =
...
+        let rec go i =
+            if i >= template.Length then
+              template.Length
+            else
+                let c = template.[i]
```
haf: Indentation

src/Logary.Facade/Facade.fs
```diff
@@ -259,31 +426,6 @@ module internal Formatting =
     let formatInstant (utcTicks : int64) =
       (DateTimeOffset(utcTicks, TimeSpan.Zero).ToString("o")) + ": "
 
-    let formatValue fields = function
```
haf: I was pretty happy with this code :-s – especially its code-size.

src/Logary.Facade/Facade.fs
```diff
@@ -315,6 +457,113 @@ module internal Formatting =
...
+    let rec go lines =
+      match exnLines.ReadLine() with
+      | null -> lines // finished reading
+      | line -> if line.StartsWith(stackFrameLinePrefix) then
+                  go (lines @ [ line, Subtext; Environment.NewLine, Text ])
+                else
+                  go (lines @ [ line, Text; Environment.NewLine, Text ])
+    go []
+
+  let literateLogLevelTokenizer = function
```
haf: How about `V`, `D`, `I`, `W`, `E` and `F` instead?

src/Logary.Facade/Facade.fs
```diff
@@ -315,6 +457,113 @@ module internal Formatting =
...
+        literateExceptionColorizer context ex
+        @ [ Environment.NewLine, Text ]
+      | _ -> [] // there is no spoon
+    let errorsExceptionParts =
+      match message.fields.TryFind "errors" with
+      | Some (:? List<obj> as exnListAsObjList) ->
+        exnListAsObjList |> List.collect (function
+          | :? exn as ex ->
+            literateExceptionColorizer context ex
+            @ [ Environment.NewLine, Text ]
```
haf: So we're doing a concat every iteration; could we do an append and List.rev in the end?

src/Logary.Facade/Facade.fs
```diff
@@ -315,6 +457,113 @@ module internal Formatting =
...
+
+  /// Split a structured message up into theme-able parts (tokens), allowing the
+  /// final output to display to a user with colours to enhance readability.
+  /// Final output sample:
+  /// "[11:00:01 INF] Hello, world! There are 100 items in your cart: [Item1, Item2, Item3]"
+  /// TODO: sample string*Token list ? too confusing?
+  let literateTokenizer (context: LiterateContext) (message: Message) : (string * LiterateToken) list =
+    let formatLocalTime (utcTicks : int64) =
+      DateTimeOffset(utcTicks, TimeSpan.Zero).LocalDateTime.ToString("HH:mm:ss", context.FormatProvider)
+      , Subtext
```
haf: Commas go on the line before.

src/Logary.Facade/Facade.fs
```diff
@@ -315,6 +457,113 @@ module internal Formatting =
...
+        Console.ForegroundColor <- color
+        Console.Write(text)
+      )
+      Console.ResetColor()
+
+/// Logs a line in a format that is great for human consumption,
+/// using console colours to enhance readability.
+/// Sample: [10:30:49 INF] User "AdamC" began the "checkout" process with 100 cart items
+type LiterateConsoleTarget(minLevel, ?context, ?colorize, ?colorWriter, ?consoleSemaphore) =
+  let sem           = defaultArg consoleSemaphore (obj())
```
haf: With multiple libraries running this, we'll have issues with this semaphore not being one any longer. We need to consider that case. Should probably be a 'global' in Logary and something that is auto-configured in the Facade adapter, right?

src/Logary.Facade/Logary.Facade.fsproj
```diff
@@ -30,21 +31,6 @@
     <WarningLevel>3</WarningLevel>
     <DocumentationFile>bin\Release\Logary.Facade.xml</DocumentationFile>
   </PropertyGroup>
-  <ItemGroup>
```
haf: Why are we moving this?

src/Logary.sln
```diff
@@ -26,103 +26,103 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "core", "core", "{EA09AD42-B
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "integration-tests", "integration-tests", "{4755795D-4488-4013-81FC-9FD5D442B211}"
 EndProject
-Project("{f2a71f9b-5d33-465a-a702-920d77279786}") = "Logary.Targets.Riemann", "targets\Logary.Targets.Riemann\Logary.Targets.Riemann.fsproj", "{CC87925C-E4F2-4386-A322-8BC088DE7B33}"
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Logary.Targets.Riemann", "targets\Logary.Targets.Riemann\Logary.Targets.Riemann.fsproj", "{CC87925C-E4F2-4386-A322-8BC088DE7B33}"
```
haf: Does this change add something?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -5,6 +5,23 @@ open Fuchu
 open Logary.Facade
 open ExpectoPatronum
 
+
+type internal TemplateToken =
+    | TextToken of string
```
haf: Indentation

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -5,6 +5,23 @@ open Fuchu
...
+    | PropToken of name:string * format:string
+    with override x.ToString() =
+                    match x with
```
haf: Indentation; things after with go on their own lines.

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -5,6 +5,23 @@ open Fuchu
...
+                    | TextToken s -> "TEXT("+s+")"
+                    | PropToken (n,f) -> "PROP("+n+":"+f+")"
+
+let internal parseTemplateTokens template =
+    let tokens = ResizeArray<TemplateToken>()
```
haf: Indentation

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
       let logger = Targets.create level
       Message.event Verbose "Hi {name}" |> logger.logSimple
 
-    testCase "extract matches" <| fun _ ->
-      let str = "Hi {ho}, we're { something going on  } special"
```
haf: Why did you erase the extra whitespace here?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
-      let strEvil = "{a}{b}{  c}{d  } {e} fghij {k}} {l} m {{}}"
-      let aThroughL =['a'..'l'] |> List.map string 
-      let fields = aThroughL |> List.map (fun s -> s, s) |> Map.ofList
+      let tokens = parseTemplateTokens template
+      Expect.equal tokens.Length 5 "Extracts 4 texts parts and 1 property part"
+      Expect.equal tokens
+                   [
+                    TextToken ("Hi ")
+                    PropToken ("ho", "##.###")
+                    TextToken (", we're ") // <- a quirk of the parser
```
haf: Isn't it a valid quirk?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
-                   "Should extract correct match values"
-
-    testCase "extract matches evil" <| fun _ ->
-      let strEvil = "{a}{b}{  c}{d  } {e} fghij {k}} {l} m {{}}"
-      let aThroughL =['a'..'l'] |> List.map string 
-      let fields = aThroughL |> List.map (fun s -> s, s) |> Map.ofList
+      let tokens = parseTemplateTokens template
+      Expect.equal tokens.Length 5 "Extracts 4 texts parts and 1 property part"
+      Expect.equal tokens
+                   [
```
haf: Lists and arrays declarations should start on the same line as the first bracket.

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+                    TextToken (", we're ") // <- a quirk of the parser
+                    TextToken ("{ something going on }")
+                    TextToken (" special")
+                   ]
+                   "Should extract correct name and format parts"
+
+    testCase "handles evil property strings correctly" <| fun _ ->
+      let template = "{a}{b}{  c}{d  } {e} fghij {k}} {l} m {{}}"
+      // validity:     +  +  --    --   +         +    +     --
+      let expectedPropertyNames = ["a"; "b"; "e"; "k"; "l"]
```
haf: You've changed the expected property names and changed the semantics here, why is that?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -99,6 +208,6 @@ let tests =
         |> Message.setSingleName "Logary.Facade.Tests"
       let subject = Formatting.defaultFormatter msg
       Expect.equal subject
-                   (sprintf "[I] %s: Hi hola, we're very special [Logary.Facade.Tests]" (nowDto.ToString("o")))
+                   (sprintf "[I] %s: Hi hola, we're { something going on  } special [Logary.Facade.Tests]\n - something going on: very" (nowDto.ToString("o")))
```
haf: Not sure about this change

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+                    TextToken ("{ something going on }")
+                    TextToken (" special")
+                   ]
+                   "Should extract correct name and format parts"
+
+    testCase "handles evil property strings correctly" <| fun _ ->
+      let template = "{a}{b}{  c}{d  } {e} fghij {k}} {l} m {{}}"
+      // validity:     +  +  --    --   +         +    +     --
+      let expectedPropertyNames = ["a"; "b"; "e"; "k"; "l"]
       let strEvilRes = "abcd e fghij k} l m {{}}"
```
haf: Can we add an assert that the formatted string equals this, please?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+      Expect.equal (subject |> List.choose (function PropToken (n,f) -> Some n | _ -> None))
+                   expectedPropertyNames
                    "Should extract relevant names"
-      Expect.equal (subject |> List.map (fun (_, _, m) -> m.Index))
-                   [0; 3; 6; 11; 17; 27; 32]
-                   "Should extract correct indicies"
 
-    testCase "format template properly" <| fun _ ->
+    testCase "literate tokenizes with field names correctly" <| fun _ ->
+      let str = "Added {item} to cart {cartId} for {loginUserId} who now has total ${cartTotal}"
```
haf: So does it matter whether you use `$` or not?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+                         "cartId", box cartId
+                         "loginUserId", box loginUserId
+                         "cartTotal", box cartTotal ]
+      let now = Global.timestamp ()
+      let nowDto = DateTimeOffset(DateTimeOffset.ticksUTC now, TimeSpan.Zero)
+      let msg = Message.event Info str |> fun m -> { m with fields = fields; timestamp = now }
+      let nowTimeString = nowDto.LocalDateTime.ToString("HH:mm:ss")
+      let context = { LiterateContext.Create() with PrintTemplateFieldNames = true }
+      let colourParts = Formatting.literateTokenizer context msg
+      Expect.equal colourParts
```
haf: Looking good

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+                    [ "[",                    Punctuation
+                      nowTimeString,          Subtext
+                      " ",                    Subtext
+                      "INF",                  LevelInfo
+                      "] ",                   Punctuation
+                      "Added ",               Text
+                      "TicTacs",              StringSymbol
+                      " to cart ",            Text
+                      "{cartId:X}",           LevelFatal
```
haf: So LevelFatal means that there's no binding for it?

src/tests/Logary.Facade.Tests/Facade.fs
```diff
@@ -62,33 +79,125 @@ let tests =
...
+                         "cartId", box cartId
+                         "loginUserId", box loginUserId
+                         "cartTotal", box cartTotal ]
+      let now = Global.timestamp ()
+      let nowDto = DateTimeOffset(DateTimeOffset.ticksUTC now, TimeSpan.Zero)
+      let msg = Message.event Info str |> fun m -> { m with fields = fields; timestamp = now }
+      let nowTimeString = nowDto.LocalDateTime.ToString("HH:mm:ss")
+      let context = { LiterateContext.Create() with PrintTemplateFieldNames = false }
+      let colourParts = Formatting.literateTokenizer context msg
+      Expect.equal colourParts
```
haf: Nice assert

src/tests/Logary.Facade.Tests/Logary.Facade.Tests.fsproj
```diff
@@ -36,34 +37,6 @@
   <PropertyGroup Condition="'$(VisualStudioVersion)' == '10.0' OR '$(VisualStudioVersion)' == '11.0'">
     <FSharpTargetsPath>$(MSBuildExtensionsPath32)\..\Microsoft SDKs\F#\3.0\Framework\v4.0\Microsoft.FSharp.Targets</FSharpTargetsPath>
   </PropertyGroup>
-  <ItemGroup>
```
haf: Do we gain something from moving this around?

src/Logary.Facade/Facade.fs
```diff
@@ -405,7 +403,7 @@ module internal Formatting =
           // but using 'LevelFatal' so it really stands out.
           themedParts.Add (prop.ToString(), LevelFatal)
 
-      FsMtParser.parseParts context.Buffer template foundText foundProp
```
haf: It's not at all a bad idea to have a context buffer. What is the reasoning here?

adamchester: Dumping the string builder results in better performance, and it wasn't thread safe either.

src/Logary.Facade/Logary.Facade.fsproj
adamchester: Done by VS, I'll remove these when I'm done, otherwise VS will just keep doing it

src/Logary.sln
adamchester: Done by VS, I'll remove these when I'm done, otherwise VS will just keep doing it

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: Damn it, this is why I tried to add editor confit at the start, but failed miserably :cry:

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: The quirk is that the open brace is potentially a property, a so gets treated as a new text token instead of a continuous single text token.

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: Because space isn't valid in a property name, is there some good reason to allow it? Logary won't allow it either, neither does: aspnet, serilog, nlog, MessageTemplates, string.format, etc

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: See other comment here: https://github.com/logary/logary/pull/155#discussion_r77165940

src/tests/Logary.Facade.Tests/Logary.Facade.Tests.fsproj
adamchester: VS :angry:

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: Oops, mistake I think

src/Logary.Facade/Facade.fs
adamchester: Maybe best to move this to the `LiterateOptions` like theme? I prefer 3 letter codes, but really it's a personal thing, I find it slightly too hard to notice the colour/level so I think it's worse readability

src/Logary.Facade/Facade.fs
adamchester: Hey it's an awesome piece of code, but regex went nuts when I tried to add format/alignment etc

adamchester: Also, to reduce parser complexity, I dropped the feature that returned the index. It's hard with support for `{{` which is better for the user who wants braces in their output

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: sorry not sure what you mean on this one, do you mean the `{ something going on  }` property?

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: It's possible I'm confused, but we don't really have a formatter method worth testing yet, do we?

src/Logary.Facade/Facade.fs
adamchester: Right, good catch. we can't have two targets writing to the console, messing up the pretty formatting.

src/Logary.Facade/Facade.fs
adamchester: That should help, I will do that, thanks.

src/Logary.Facade/Facade.fs
haf: Ah, I see your point. Let me think about it, but if I don't comment again, feel free to keep the current version.

src/tests/Logary.Facade.Tests/Facade.fs
haf: Yes, that one.

src/tests/Logary.Facade.Tests/Facade.fs
haf: But space isn't a valid property name in the previous implementation either; it's just spacing to make it read better. I think people will get confused by not being allowed to space out around the text. It's the trimmed non-space-character text that equals the property name.

And spaces *within* text surely can be proper property names?

src/tests/Logary.Facade.Tests/Facade.fs
haf: Doesn't coloured console format it? This is a lower prio from my side, it's mostly a question.

src/Logary.Facade/Facade.fs
haf: Btw, do move it.

haf:
Generally; is this PR equivalent to the coloured console target at this stage?

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: No spaces anywhere in template property names. The following similar libraries with similar template syntax do not allow any space, anywhere, in the property name: [aspnet logging](https://github.com/aspnet/Logging/blob/dev/samples/SampleApp/Program.cs#L113), serilog, message templates, string.format, [nlog](https://github.com/NLog/yamtp)

So, I would only say for the sake of compatibility, we should avoid spaces inside template property tags.

adamchester: When I say compatibility, I'm thinking if you allow spaces, and some target doesn't allow spaces, something has to be done about it. If you prevent spaces, nothing can possibly go wrong

adamchester:
I think it's more than just colour; it's about human readable output (eg. Like printf or string.format support .net format strings).

But yeah, colour is a big part of making it easy to scan the log output for key information.

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: No, in this case, $ was part of text, not part of the property. One could make a fancy .net format string for currency though, maybe I should add a test for that.

haf: Let's keep the features down.

adamchester: I agree on keeping features down, but this one you can already do, and it's a reasonably big part of this is PR - for readability sake - the ability to allow standard .NET format strings, e.g. `You have {percent:##.##) available on {date:dd-MMM-yyyy}`. this is allowed by the parser, and simply formats via `String.Format` with an `IFormatProvider`.

adamchester: note: i hit this in Suave a few days ago, needing to format the [log message dates](https://github.com/SuaveIO/suave/commit/b49afc42de70227ee38f742ab24be61c48e80929#diff-24150220eb6eb90c91d197d269bb15c5R364), and [started milliseconds](https://github.com/SuaveIO/suave/pull/499/files#diff-6f6ce16f18371e4ccd72aa620e83dec5R262).

haf: Yes, I saw that. I tweeted from SuaveIO about this being the 'backing' code for the new feature – isn't that the case?

haf:
Ok, tell me when you need another review. Starting to look good!

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: I hope so ;) yes

src/Logary.Facade/Facade.fs
adamchester: That's the signal that we're finished (reached the end of the template string).

src/Logary.Facade/Facade.fs
adamchester: Not sure what to do about this, and should we do it now or after this PR?

src/tests/Logary.Facade.Tests/Facade.fs
adamchester: now it's `MissingTemplateField`

src/Logary.Facade/Facade.fs
adamchester: Hopefully the comment covers it now.

adamchester:
OK, obviously since I made such a formatting mess (sorry!) there was a lot to fixup, but I think I've caught them all now, except for:

 1. Global / console semaphore - now or after?
 1.  Still trying to find a nice way to concat the token parts together without using `@` (concat)

It's probably ready for another review. Once you're happy, I'll remove the other garbage (VS solution changes, editorconfig, etc).

haf:
Semaphore... After I added the defaulting semaphore to the console target that comes with the logging, I added [the global semaphore](https://github.com/adamchester/logary/blob/3b16a70830e723e7cc20d95beada4234708f0a3d/src/Logary.Facade/Facade.fs#L712)

But that semaphore already has a purpose.

What do you think of adding a consoleSemaphore to the config and having the literate printer as the default target?

haf:
As for point 2 – there's nothing wrong with the concat operator as long as it's not used in a loop over many items. Keep it as it is now; your `List.rev` changes are looking good.

haf:
Another question; do you know of anything you've added that may preclude a consuming library for compiling with .Net Core?

haf:
PRed to your PR to show what I mean... Can you fix the surrounding API?

adamchester:
I doubt anything will prevent netcore, culture stuff is there, console colour is there, string.format is there. but I can check if you like? I would also like it to be netcore compilable asap.

I was thinking of dropping all the list concats in favour of a less GC damaging ResizeArray or something. However, on the other hand, this target doesn't really need to perform that great, because if you have thousands of lines scrolling past quickly, you're using the wrong target :+1: 

Re:semaphore, when I'm back at a computer I'll have another look at it :+1:

haf:
Ready to merge? Did you test the Facade example and run the units tests?

adamchester:
@haf could you please take a look at https://github.com/logary/logary/pull/155/commits/42d5afd080b06b59ac76a4f3d68886fc4a90f493#diff-decd44de4c55ff3c7ecc144865ae3b38R199 and tell me if you have a plan for where the semaphore originates in this case ?

src/adapters/Logary.Adapters.Facade/Logary.Adapters.Facade.fs
```diff
@@ -195,6 +195,8 @@ module LogaryFacadeAdapter =
             field.PropertyType,
             fun _ -> Date.timestamp () |> box)
 
+        | "consoleSemaphore" ->
+          obj() // Todo: should this come from Logary now? where?
```
haf: From https://github.com/logary/logary/blob/master/src/Logary/Internals_Globals.fs#L8 as a new object. Where's Logary's official colourised target btw? I was gonna point to it but ... ;)

adamchester:
I was thinking, we should provide a really basic textwriter target too, because we are removing the ability to have the log format that previously available (but the code behind it is still here). Also, with textwriter, it can be much faster than the slow console colour logging.

src/adapters/Logary.Adapters.Facade/Logary.Adapters.Facade.fs
adamchester: Hah, yeah Logary proper should get the same ;)

haf:
@adamchester Thanks for your awesome work. A PR to clone the colourised target to Logary proper would be nice too ;)

adamchester:
@haf I'll see what I can do. I started experimenting with it.

You probably want to fix the console lock btw :blush:

PR #156, opened 2016-09-01
Feature/improved supervisor

mavnn:
This still needs some more unit testing on the new supervision code, but I'm starting the pull request so you can see where things are heading.


src/Logary/Target.fs
```diff
@@ -108,6 +103,19 @@ module TargetUtils =
...
         let! requests = RingBuffer.create 500u
         let shutdown = Ch ()
+        return { server   = fun _ _ -> loop metadata requests shutdown
+                 requests = requests
+                 shutdown = shutdown
+                 name     = name' }
+      }
+    }
+
+  let willAwareNamedTarget loop name : TargetConf =
```
mavnn: @haf this is where I added a new method for registering "willAware" targets so existing targets can continue to work unaltered. If you can think of a better name, let me know...

haf: @mavnn willful? ;)

haf:
Will you have a look at @neoeinstein 's work too, @mavnn ?

mavnn:
Yes, few nice looking tweeks I can steal

PR #157, opened 2016-09-01
Allow dotted names in message templates

neoeinstein:
Fields in a log message may use dotted names. This allows the template engine to access those names.

Possible future changes may include using dots to reach into fields with object values to access child items.

### DIFF
--- a/src/Logary/MessageTemplates.fs
+++ b/src/Logary/MessageTemplates.fs
@@ -157,7 +157,7 @@ module Parser =
         else startAt, Empty.textToken
 
     let inline isLetterOrDigit c = System.Char.IsLetterOrDigit c
-    let inline isValidInPropName c = c = '_' || System.Char.IsLetterOrDigit c
+    let inline isValidInPropName c = c = '_' || c = '.' || System.Char.IsLetterOrDigit c
     let inline isValidInDestrHint c = c = '@' || c = '$'
     let inline isValidInAlignment c = c = '-' || System.Char.IsDigit c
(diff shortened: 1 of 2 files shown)


haf:
/cc @adamchester 

Also see elmah.io's formatting w/ dots and how interaction with it may be

adamchester:
How about `.h` or `h.`, are they valid? how will popular targets treat behave if they were encountered?

/cc @nblumhardt - does Seq allow dotted names? obviously you chose originally to prevent dots in Serilog (which seem to me a really common thing to do), did you happen to remember any reason why?

haf:
Elmah is here https://github.com/logary/logary#what-does-it-look-like

nblumhardt:
@adamchester the issue is the interpretation of:

```csharp
Log.Information("Logged on {User.Name}", "Adam");

// vs

Log.Information("Logged on {@User}", new { Name = "Adam" });
```

Is the first one really a property name with a dot in it, or is it shorthand for a property called `User` with a single member `Name`?

I think that the logical intent is generally the second one - the dot's not part of the property name.

Likewise in the case of elmah.io, the dotted property names are just a way of serializing the nested structure - so the second example is the "semantic" way of sending nested structures through Serilog; the sink should just be able to flatted out a `User` structure into a list of dotted property-path/value pairs.

I'd really like to be able to support example 1 in Serilog as e.g.:

```csharp
Log.Information("Logged on {@User.Name}", "Adam");
```

This is actually where the term "destructuring" (as in destructuring assignment) actually came into the lexicon :-)

I'd recommend supporting the elmah case through the structured data pipeline, rather than introducing dotted property names, unless you're really certain that the semantics are correct.

HTH!

PR #158, opened 2016-09-01
Fix level required to log for `logDebug`

neoeinstein:
`logDebug` was requiring that a target accept logging at the `Verbose` level in order to forward messages.

For `fMessage`s that were consuming the `LogLevel` passed in, they were also being incorrectly set with the `Verbose` log level.


haf:
Thank you!

PR #165, opened 2016-09-19
[WIP] Literate console target

adamchester:
No need to review yet - it's just a start 👍

### DIFF
--- a/src/Logary/Targets_Core.fs
+++ b/src/Logary/Targets_Core.fs
@@ -117,25 +117,27 @@ module LiterateConsole =
   open Logary.Internals
   open Hopac
 
-  /// The output tokens, which can be potentially coloured.
-  type LiterateToken =
-    | Text | Subtext
-    | Punctuation
-    | LevelVerbose | LevelDebug | LevelInfo | LevelWarning | LevelError | LevelFatal
-    | KeywordSymbol | NumericSymbol | StringSymbol | OtherSymbol | NameSymbol
-    | MissingTemplateField
+  module Tokens =
+    /// The output tokens, which can be potentially coloured.
+    type LiterateToken =
+      | Text | Subtext
+      | Punctuation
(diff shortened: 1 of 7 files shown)


adamchester:
image

src/Logary/Targets_Core.fs
```diff
@@ -108,6 +108,234 @@ module TextWriter =
...
+      match value with
+      | Value.Array items -> [ "todo", OtherSymbol ] // TODO:
+      | Value.String s -> [ s, StringSymbol ]
+      | Value.Bool b -> [ b.ToString(), KeywordSymbol ]
+      | Value.Float f -> [ f.ToString(), NumericSymbol ]
+      | Value.Int64 i -> [ i.ToString(), NumericSymbol ]
+      | Value.BigInt bi -> [ bi.ToString(), NumericSymbol ]
+      | Value.Binary (x, y) -> [ "todo", OtherSymbol ] // TODO:
+      | Value.Fraction (x, y) -> [ "todo", OtherSymbol ] // TODO:
+      | Value.Object o -> [ o.ToString(), OtherSymbol ] // TODO: recurse
```
haf: In the `Logary.Formatting` module there's some logic to traverse to form keys like `a.b.c => 3, a.b.d => 55`; the above function would be a nice parameter into that function to avoid having to recurse manually here.

adamchester:
Another WIP screenshot 
image

haf:
Starting to look good :). Is that on Windows?

adamchester:
It is Windows, so I could use the perf counters. I hacked the existing sample to get those. Will consider a separate sample when things stabilise ;)

adamchester:
One interesting issue I found is what to do with invalid format strings. I wonder if should try to render without format string in those cases, do you have an opinion?

haf:
Yes, doing a verbatim print is probably useful then.

adamchester:
I'm goign to try to split some of this up into smaller, more targeted, separate PRs.

PR #170, opened 2016-09-27
[WIP] Literate console target

adamchester:
Work in progress, no need for reviews yet 👍

### DIFF
--- a/examples/Servizz/Program.fs
+++ b/examples/Servizz/Program.fs
@@ -10,7 +10,10 @@ open System.Threading
 
 let logger = Logging.getLoggerByName "Servizz.Program"
 
-type MyDu = UserLoggedIn of name:string | UserLoggedOut of name:string
+type Speed = Fast | Slow | Other of speed:int
+type Event =
+| UserLoggedIn of name:string * roles:string list
+| UserLoggedOut of name:string * speedAtTimeOfLeaving: Speed
 
 [<EntryPoint>]
 let main argv =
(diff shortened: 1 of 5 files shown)


adamchester:
Updated exception new lines, and started on object (map) rendering
img_0188

adamchester:
image

adamchester:
image

adamchester:
- [ ] Fix exception prefix on Mono

image

examples/Servizz/Program.fs
```diff
@@ -10,14 +10,23 @@ open System.Threading
 
 let logger = Logging.getLoggerByName "Servizz.Program"
 
+type Speed = Fast | Slow | OtherSpeed of speed:int
+type LoginLogoutEvent =
+| UserLoggedIn of user : string * roles : string list
```
haf: 2 space indentation

src/Logary/TemplateFormat.fs
```diff
@@ -0,0 +1,101 @@
+﻿[<AutoOpen>]
+module TemplateFormat
```
haf: Why not make this a single module, like so: `[<AutoOpen>] module internal Logary.LogaryCapturing`?

src/Logary/TemplateFormat.fs
```diff
@@ -0,0 +1,101 @@
...
+
+  let logyarDefaultDestructurer : Destructurer =
+    Capturing.createCustomDestructurer (None) (Some Capturing.destructureFSharpTypes)
+
+  let captureField (property : Utils.FsMessageTemplates.Property) value =
+    logyarDefaultDestructurer (DestructureRequest (logyarDefaultDestructurer, value, 10, 0, property.Destr))
+    |> convertTemplatePropertyToField
+    |> fun f -> Field (f, None)
+
+type Message with
```
haf: @mavnn Compared to *Xenogloss*, what do you think of this? I think it's a good trade-off in that it doesn't add any dependencies; but there may be easy gain to be had by adding some of your code, perhaps?

adamchester:
Fixed exception stack trace prefix on mono:
image

haf:
This seems ready to merge for me :) – do you agree?

adamchester:
I have a whole bunch of tests to add yet, and potentially some bits I haven't considered fully yet. I'm OK with merge & fix later, in smaller increments, if you think that's wise.

haf-lynx:
Especially tests are good, but I think this is good enough, since the coloured target is currently opt-in rather than default. I'm also considering giving you commit access to the repo; what would you say about that? :)

adamchester:
Ok 👍

PR #183, opened 2016-09-29
Less new lines in literate exception rendering

adamchester:
After this change, there are far fewer new lines when exceptions printed.

This screenshot shows what it looks like *after*:
image


haf:
Nice. But do we want the final newline? Or the one before the inner exception?

adamchester:
That's a good question, I'll figure out why that's happening

adamchester:
Updated:
image

haf:
Thanks!

PR #194, opened 2016-10-12
[facade] fix exn stack line prefix on mono

adamchester:
after:
image

before, the stack trace lines were brighter.

This now matches the behaviour of Logary proper.


haf:
Thanks @adamchester !

PR #219, opened 2016-12-13
WIP: support log event stream processing and format improvement

haf:
- remove pollServices in registry

  Because supervisor will supervise minions, if an exception occurs, it will be handled according to policy. So there's no need for state polling, because we can't do much if we can get the status. the user's handle logic can be expressed by policy.

- remove healthcheck & metric

  >  The aim here is to normalise the Gauge/Derived/Message semantics into a single semantic of 'things computed from events from your app'.

  - remove healthcheck

    Looks like metric can also be seen as a healthcheck, such as windows performance counters. when outside tick , metric generate some msg, which will be processed by user defined engine , they can decide their meaning (log or metric or healthcheck...) there.

  - remove metric

    metric can be implemented by engine's process expression builder, what we need to do is schedule jobs for tick, in the corresponding time(like old metricConf.tickInterval) or manually tick. Metric's data sources can have two types, one is similar to WPC, generate data in real-time when ticked .the other is generate through the normal log message.

- support multi targets on internallogger and apply rules when send msg to target

- redefined Message

  The aim here is to normalise the Gauge/Derived/Message semantics into a single semantic of 'things computed from events from your app'.

  - remove FieldModule.fs
  - use FsMessageTemplate for default MessageWriter
  - remove Chiron json support for message type 
  - no need SuppressPointValue
  - use obj instead of `Value Modle`
  -  gauges are stored in contexts with gauge type. so message.value is template or raw message.

- errors (exceptions) 

  is expressed by a context, not a field. user can define their own output template for decision whether or not to show them. 

- formatting
 
   contexts other than fields in template will destructure with `@` in default message writer

- todo
  - [x] more unit test
    - [x] message
    - [x] ewma & percentile
    - [x] formatting
      - [x] migrate old formatting tests
      - [x] consider msg with gauges
      - [x] consider msg with gauges, fields, contexts
    - [x] migrate old core targets tests
    - [x] registry
  - [ ] migrate whole solution **help wanted**
    - [x] facade
    - [x] targets
    - [ ] SuaveReporter
    - [ ] Targets.InfluxDb
    - [x] examples with new style
  - [x] event format improvement mentioned by #257
  - [x] reorganize types in Events.fs
  - [ ] Support building for .Net Core and .Net FW

  - to be discussed
    -  do we need both `logWithAck` and `log` api in `Logger` type,beacuse `log` is just delegate call on `logWithAck` and ignore its result (`promise`)
    - registry/logmanager IDisposable ?
    - csharp factory support processing
    - new healthcheck/metric style


lust4life:
image


@haf   hi, what ide/editor do you use, i open `sln` in vs2017 , but can't open `logary.fsproj` (load failed), maybe because  vs can not support `dir` files split ?

haf:
Hi, I'm using VSCode. Sorry about the problems you're experiencing with VS. Perhaps it's lacking a `<Folder ... />` and requires `\` instead of `/`?

lust4life:
```cmd


"E:\git\logary\logary\src\Logary.sln" (default target) (1) ->
"E:\git\logary\logary\src\targets\Logary.Targets.Riemann\Logary.Targets.Riemann.fsproj" (default target) (2) ->
"E:\git\logary\logary\src\Logary\Logary.fsproj" (default target) (3) ->
(CoreCompile target) ->
  E:\git\logary\logary\src\Logary\Metric.fs(198,7): error FS0001: This expression was expected to have type↔    'Iterable'    ↔but here has type↔    'unit' [E:\git\logary\logary\src\Lo gary\Logary.fsproj]
  E:\git\logary\logary\src\Logary\Registry.fs(73,47): error FS0001: The type 'NodeInput' is not compatible with the type 'Alt<NodeInput>' [E:\git\logary\logary\src\Logary\Logary.fsproj ]


"E:\git\logary\logary\src\Logary.sln" (default target) (1) ->
"E:\git\logary\logary\src\tests\Logary.Adapters.Facade.Tests\Logary.Adapters.Facade.Tests.fsproj" (default target) (39) ->
"E:\git\logary\logary\examples\Libryy\Libryy.fsproj" (default target) (40) ->
  FSC : error FS1225: File 'E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.optdata' not found alongside FSharp.Core. File expected in E:\git\logary\logary\examples\Libryy\b in\Debug\FSharp.Core.optdata. [E:\git\logary\logary\examples\Libryy\Libryy.fsproj]
  FSC : error FS0229: Error opening binary file 'E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.dll': File 'E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.optda ta' not found alongside FSharp.Core. File expected in E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.optdata. [E:\git\logary\logary\examples\Libryy\Libryy.fsproj]
  FSC : error FS3160: Problem reading assembly 'E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.dll': Error opening binary file 'E:\git\logary\logary\examples\Libryy\bin\Deb ug\FSharp.Core.dll': File 'E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.optdata' not found alongside FSharp.Core. File expected in E:\git\logary\logary\examples\Libryy\bi n\Debug\FSharp.Core.optdata. [E:\git\logary\logary\examples\Libryy\Libryy.fsproj]
  error FS0073 : internal error : BuildFrameworkTcImports: no successful import of E:\git\logary\logary\examples\Libryy\bin\Debug\FSharp.Core.dll [E:\git\logary\logary\examples\Libryy\ Libryy.fsproj]

    2 Warning(s)
    6 Error(s)
```

@haf   can this branch build success ? it show above message on my machine, did i miss someting? or is beacuse this is WIP  ?

haf:
In this branch, you should only be able to build Logary.fsproj and Logary.Tests.fsproj; the rest are WIP. I normally use `xbuild src/tests/Logary.Tests/Logary.Tests.fsproj && mono --debug src/tests/Logary.Tests/bin/Debug/Logary.Tests.exe --summary --debug`

lust4life:
any plan on this pr , seems big rewrite.

code & structure are more clearer (good job), but some unfinished code make me little frustrating 😑

haf:
Hi there,

Yeah, it's a big rewrite of the internals, but thankfully almost nothing of the public API has actually changed so it will be easy for consumers to upgrade. For targets; medium difficulty.

The aim is to port something of Nessos Streams into a metrics engine, so that the configuration of where to send what names/hieras/messages/metrics/etc is done with code rather than the rules. That way we can add in calculating metrics, like finding the 99th percentile of request latency every rolling 5 second window.

Logary is very integral to our operations, so I will indeed be finishing this PR. As far as I remember, "all" that's left is the actual "engine" part of things, i.e. the code that will tie filters, mappers, reducers, metrics and targets together.

I have some diagrams I could post, if that would help?

Like always I would love to have someone like yourself to discuss this rewrite with, especially the fun part that's all that's left; the engine.

src/Logary/Targets/Noop.fs
```diff
@@ -81,7 +82,10 @@ module internal Impl =
 
 /// Create a new YOUR TARGET NAME HERE target
 [<CompiledName "Create">]
-let create conf name = TargetUtils.stdNamedTarget (Impl.loop conf) name
+let create conf name =
+  TargetConf.create Policy.exponentialBackoffSix 500u (Impl.loop conf) name
```
haf: Targets specify their own default policies.

src/Logary/Internals/Service.fs
```diff
@@ -0,0 +1,115 @@
...
+             (server : SupervisedJob<unit>)
+             : Job<Service<T>> =
+    let ilogger = ilogger |> Logger.apply (setSimpleName (sprintf "Logary.Service(%s)" name))
+    let pn = PointName.ofSingle name
+    let mserver = memo server
+    let pauseCh, resumeCh, getStateCh = Ch (), Ch (), Ch ()
+
+    let rec running transitions =
+      Alt.choose [
+        pauseCh ^=>. pause
```
lust4life: https://github.com/logary/logary/pull/219/files#diff-80a43381430d2c2d495b79e45685db17R48

does the `pause` same with `pauseCh` ?  confusing `Alt.choos` in `running` method here

src/Logary/Targets/Noop.fs
lust4life: get this idea, every service can have its own policy for supervisor ?

src/Logary/ValueModule.fs
```diff
@@ -0,0 +1,252 @@
...
+          yield "data", valueOfObject e.Data ]
+
+    HashMap.ofList <|
+      if e.InnerException <> null then
+        ("inner", Object <| exceptionToStringValueMap valueOfObject e.InnerException) :: fields
+      else
+        fields
+
+  // TODO: move out of this file,
+  // TODO: use TypeShape,
```
lust4life: how can typeshape help here  ?
```fsharp
  let createUnit (value : 'a) (units : Units) =
    failwith "TODO: TypeShape"

  let create (value : 'a) =
    failwith "TODO: TypeShape"
```

haf: It caches the boxing + reflection and makes it possible to create values configurable: by default we should call `ToString()` on them, but allow the user to add more `Visitor`s to the configuration.

src/Logary/Internals/Service.fs
haf: `pause` is the function, `pauseCh` is the value channel.

lust4life:
> Hi there,
>
>Yeah, it's a big rewrite of the internals, but thankfully very little of the public API has actually changed so it 

thanks for describing the design purpose , help building the rough big picture in mind. but the scattered details and my skills (need have a really look at hopac )  make me crashed 😶  , will continue keep an eye on this pr, and give what i can support within my ability .

and the `retry/reconfig` feature is really useful , e.g. 
- generally we set some level (error/warn) for logging on prod envir , but when their is some problem need to be fixed online , we want to lower the level on the fly for some mins in order to show the info or verbose msg to help locate/analyse the problem. 

seems can be done by invoke some reconfig api through the caller/application site.

and the new design seems more flexible and robust. good job 👍  , hopefully it will move forward.

haf:
If you're interested in helping out but are feeling stuck on Hopac, I could give you a tutorial in return for help to complete this branch.

I suppose you've seen https://github.com/Hopac/Hopac/blob/master/Docs/Programming.md and https://haf.github.io/2017/04/13/ThreadSafeDictionary-in-Hopac-in-fsharp.html too?

lust4life:
after read these links a few more times again, the basic usage of hopac is more clearly now, thx . 

lack of some context about the `Engine` part and `Processing` , can't really imagine the usage.
how about give more test case in `Engine.fs` ? 

```fsharp
    testCase "play" <| fun () ->
      let processing =
        Events.stream
        |> Events.subscribers [
          Events.filterBy (service "riak")
          |> Events.sink "email" ["henrik@haf.se"]

          Events.filterBy (service "spns")
          |> Events.sink "elasticsearch"
        ]
        |> Events.toProcessing

      Engine.create processing |> ignore
```

is there any context about the `Events` api ？ is it the `Flow<'t>` extension method in `metric.fs` ?

src/Logary/Internals/Service.fs
lust4life:
```fsharp
module internal GlobalService =
  open Global

  let create (t : T) (ilogger : Logger) =
    let logger = ilogger |> Logger.apply (setSimpleName "Logary.Globals")
    let pauseCh, resumeCh, shutdownCh = Ch (), Ch (), Ch ()

    let rec init () =
      let prev = !config
      initialise t
      running t (fst prev)

    and running myself prev =
      Alt.choose [
        pauseCh ^=> fun (ack, nack) ->
          logger.debug (eventX "Pausing.")
          initialise prev
          ack *<= () >>=. running myself prev

        resumeCh ^=> fun (ack, nack) ->
          logger.debug (eventX "Resuming.")
          initialise myself
          ack *<= () >>=. running myself prev

        shutdownCh ^=> fun ack ->
          logger.debug (eventX "Shutting down.")
          initialise prev
          ack *<= ()
      ]

    let shutdown : Alt<Promise<unit>> =
      shutdownCh *<-=>= fun repl -> repl
      |> Alt.afterFun (fun iv -> iv :> _)

    let loop = Job.supervise logger Policy.terminate (init ())
    Service.create logger "globals" pauseCh resumeCh shutdown loop
```

if the `pause` is the function , the last line above passing an `channel` to it, is this work in progress，just for temporary compile ?

haf:
Yeah, the Events API is supposed to mimick Riemann.io so that you can compose nested expressions like https://stackoverflow.com/q/38483416/63621

`sink` will be a simple string and usages of it will "inspect" the available targets and find one with a matching name.

The idea is that both reservoirs, filters, distributions etc are exposed in this Processing expression.

lust4life:
@haf  hi , take some look at  Riemann's stream, seems what we need is a push based stream processing. 
`Nessos Streams` is implement by an iterator design pattern, seems pull based and synchronized(on demand). and `Stream` in hopac is pull based too. 

log message from logging api is asynchronous. so , is `Rx` here more suitable for `reservoirs, filters, distributions...` ? 

what we need to do is proxy logging msg (log,metric,healthcheck) by Engine, then use `Rx` method composite some async data processing on message. after processing, push the result on target buffer if the message(result) is marked to some target by sink.

haf:
I don't think Nessos Streams push vs pull properties is an issue, the composition work just needs to be done. Even if Riemann invokes directly into the next function, Streams do something similar, even if this is triggered from the outside in a _pull based fashion._

The reason I chose those is that we don't need to do anything async. Instead we have the sinks that are references to our targets, such that if a message ends up there, it's placed in a circular buffer and fed to the asynchronous target. The circular buffer is the push-back.

lust4life:
i have some questions about the pull based fashion. can you help explain some ?

if we choose pull based ,  where is the place we pull data ? use a `server job` in the `engine` loop forever invoke the `tryAdvance` from Iterator ? but the tryAdvance is a synchronize method， does this block the current thread ?

and how to express a unbound source in `Flow<‘T>` , beacuse we don't know when log message will come and seems it never end (unless the application shutdow logary), so really can't imagine how to implement code below ?

```fsharp
  let ofStream (source:Src<'T>) : Flow<'T> =
    Flow (fun { complete = complete; cont = iterf; cts = cts } -> 
      let bulk () = 
        let stream = Src.tap source
        Stream.iterFun iterf stream |> start
      let iterator() = 
        let stream = Src.tap source
        {
          new Iterator with
            member __.dispose () = ()
            member __.tryAdvance () = 
              // how to implement here ?
              true
        }

      {
        new Iterable with
          member __.iterator = iterator()
          member __.bulk () = bulk ()
      }
    )

```

haf:
I wasn't planning on creating flows from streams; the engine is called by the caller, and should have no particular buffering, because I assume that the contention point is the asynchronous target, and the I/O done there, rather than the CPU-bound computation.

So the signature of log is something like `LogLevel -> Message -> Alt<Promise<unit>>`; it's primarily the inner Promise we care about in order to do back-pressure of logging (when I/O cannot keep up), so we can assume the outer `Alt` is the mechanism for transferring and evaluating the log message.

Then, if you can build something that can be *synchronously* dispatched into, that is, normally invoked as a function, as input to the engine; all of the engine's execution goes into the `Alt.prepareFun`'s preparation phase. After the engine executing; this is what I want build in a synchronous manner with Flow, we either throw the message away or we drop it into a RingBuffer.

So the task initially is to build the engine in the most straight-forward manner, and then we will be adding orthogonality at the edges of it. That is; exactly what you're asking about, but later, when we have a design that can compute. The aim here is to normalise the Gauge/Derived/Message semantics into a single semantic of 'things computed from events from your app'.

lust4life:
@haf  send a pr #263  for concrete discussion. will you take a look at it ?

lust4life:
> Polling of services
> I'm also considering what to do when services; part of Logary, fail; such as a let-it-crash target. 

looks like we already have `supervise` and `policy` , whether they're used to deal with such scenes ?

haf:
Yes, I'm for reusing `supervise` and `policy` rather than a new `Service` concept.

I haven't had time to review your changes other than briefly, but I need to spend some time with the code to give good feedback.

lust4life:
Ok, I'll try to fix the rest in a simple way, so that the branch can work first.

lust4life:
Should we remove the healthcheck? Looks like metric can also be seen as a healthcheck, such as windows performance counters ? 
when outside tick , metric generate some msg, which will be processed by user defined engine , they can decide their meaning (log or metric or healthcheck...) there.

if so , i think we should bring `Scheduling` back.

haf:
I'm reluctant to remove HealthCheck, but it needs a better public API. Let's just leave it there for now as a reminder.

Yeah, we need sources of values which can be fed into the engine, like the ticking of metrics.

lust4life:
@haf  ping, do you have some time to take a look at this pr ? `Registry` `Targets` can work now, the most part of `Formatting` is already finished, but still need to deal with some corner case. i will push it forward, but there are some parts (some targets) which i am not very familiar with, and for `json formatting`, I'd like to discuss it with you.

lust4life:
plan : 
- adding a few more unit tests for `Formatting`
- maybe migrate old `CoreTarget` tests
- first make `WPC` build sucess
- then make the whole repo build success
  but since lack of more tests, maybe have some risk.

lust4life:
@haf , most parts of the whole solution can build sucess now, can you help fix some parts which i'm not familiar with?  After migration with some added tests, i think we can merge with master branch and then push it to the next version.

lust4life:
@haf  All the solution can build success now, the breaking changes are comments and use `failwith "TODO: needs to be discussed"` (12 about targets, 1 about `SuaveReport` service, 1 about `CommonLogging.Adapter`) . so the tests around with them will failed. 

make a issue about this discussion #275

lust4life:
@haf  try to fix the ci test suite, in `AppVeyor` it shows :
```
[05:49:45 ERR] logstash/start and stop errored in 00:00:00.0730000 <Expecto>
System.Exception: Exception raised by job ---> System.DllNotFoundException: Unable to load DLL 'libzmq': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

i look back to some history successful builds, found that they haven't this problem, seems the `logstash.tests.exe` doesn't run. did i miss something ? on my machine, i copy these `dll` from https://github.com/zeromq/fszmq/tree/master/lib/zeromq/WIN here, which can pass tests. but i don't know how to make it flexible in the `Rakefile` that can work both at win & linux. can you give some advice ?

beyond this, i think the test suite can pass now.

haf:
@lust4life In that case the "right" solution is to do a conditional copy if the OS is Windows-something, of "libzmq.dll" to the output folder.

An easier way forward is to copy it there with the appveyor yaml.

haf:
@lust4life How about we meet up over Skype and summarise/plan the release of this PR? I'm available evenings (UTC+1) and weekends (right now).

haf:
Could you take this travis file? https://github.com/haf/expecto/blob/master/.travis.yml

lust4life:
travis file has been updated.

haf:
Just a single test failure now

[14:27:05 ERR] logary/Engine/health checker ping failed in 00:00:03.0820000. 
ping github. Actual value was true but had expected it to be false.

lust4life:
I just skip this test, since this will use a real ping when test, maybe travis environment can not support ping outside?  health checker here is just for making an example (show how to use the new style to implement a health checker, users can port old style to this new style). so use ptestcase here.

haf:
Skip is fine.

haf:
So, now we're passing :) 👍

lust4life:
@haf  update something in readme, i think it is enough to give instruction for using this version. WDYT.

src/Logary/Configuration/Factory.fs
```diff
@@ -95,6 +92,11 @@ and ConfBuilder(conf) =
     |> Config.processing processor
     |> ConfBuilder
 
+  member x.LoggerMinLevel(path: string, minLevel: LogLevel) =
```
haf: lowercase initial letter

haf:
https://www.nuget.org/packages/aliyun-log-csharp-sdk/ needs upgrading to .net core

haf:
Now, let's see what SuaveReporter can be fidxed with

haf:
Should be fairly stable to build now; yet to run all tests and fix suave reporter and influx, but I did set up a docker compose that runs influxdb locally on docker, so the distance to go there should be pretty small.

haf:
Builds on CI now, failing on tests due to ordering errors of new TypeShape file.

lust4life:
@haf nice, glad you come back to push forward. will follow up in later days,a little busy these two days .

haf:
screen shot 2018-03-17 at 15 31 16


What do you think of this?

We need some mechanism for ignoring certain properties from the JSON input, too.

PR #222, opened 2016-12-30
Stackdriver Target v2

baronfel:
I started actually using the stackdriver target and noticed that it wasn't working at all.  I debugged it and realized that it was because I was incorrectly doing the batch request handling, and I didn't understand what I was doing with Hopac.  So until I learn more, this PR reverts the target to handling single messages at a time.  In addition, I've taken the chance to make the messages written to stackdriver much more in line with what consumers of that platform expect.

There's a concept of `labels` to a message that match well to context, so I pulled that data from the overall payload and brought it up into the labels field.

I have one question before ending this PR:  I'd like to render the templated form of the event message, but haven't yet seen how to do that. Can you offer guidance?

- Before Minimized
screen shot 2016-12-29 at 7 10 02 pm

- Before Maximized
screen shot 2016-12-29 at 7 10 20 pm

- After Minimized
screen shot 2016-12-29 at 7 10 09 pm

-After Maximized
screen shot 2016-12-29 at 7 10 40 pm


src/targets/Logary.Targets.Stackdriver/Target_Stackdriver.fs
```diff
@@ -64,15 +69,50 @@ module internal Impl =
...
-    
-    let overrides = Map.add "name" (String <| string m.name) values
-    let json = Google.Protobuf.WellKnownTypes.Struct.Parser.ParseJson (Json.format <| Object overrides)
-    LogEntry(Severity = toSeverity m.level, JsonPayload = json)
+    // labels are used in stackdriver for classification, so lets put context there.
+    // they expect only simple values really in the labels, so it's ok to just stringify them
+    let labels = m.context |> Map.map (fun k v -> Value.toString v) |> Map.toSeq |> dict
+
+    let message = 
+      // TODO: render the template with field values
```
baronfel: here is where I'd like to format the message

src/targets/Logary.Targets.Stackdriver/Target_Stackdriver.fs
```diff
@@ -87,7 +127,7 @@ module internal Impl =
     | ComputeInstance(zone, instance) -> 
         [ "project_id", project
           "zone", zone
-          "instanceId", instance ]|> dict
+          "instance_id", instance ]|> dict
```
baronfel: typo fix here to prevent exceptions from the client.

src/tests/Logary.Targets.Stackdriver.Tests/Program.fs
```diff
@@ -56,27 +56,19 @@ let target =
         |> Stackdriver.Impl.write
 
       Expect.equal subject.Severity LogSeverity.Warning "severity should be warning"
-      Expect.equal (subject.JsonPayload.Fields.["context"]) (Google.Protobuf.WellKnownTypes.Value.ForStruct(Struct.Parser.ParseJson("""{ "service" : "tests" }"""))) "should have correct context"
+      Expect.equal (subject.Labels.["service"]) "tests" "should have correct context"
       
     testCase "send" <| fun _ ->
+      let conf = Stackdriver.create targConf "test"
```
baronfel: I get repeated failures when I try to make this a proper test, by querying the count of messages posted, so I've had to neuter it.

src/targets/Logary.Targets.Stackdriver/Target_Stackdriver.fs
adamchester: How about [this](https://github.com/logary/logary/blob/06a0ded5bcc172649e08110e2409de502c3612ac/src/Logary/Formatting.fs#L17)

adamchester: Also, it's probably still a good idea to send the template right? Without that, how would you group related log events?

baronfel: I could send the template still, you're right.  But this `message` property is given special treatment in the UI (as shown in the after-collapsed view above) so I want it to be glance-able.

baronfel: Thanks, that ended up working perfectly!

haf:
Seems like you found the templating.

baronfel:
@haf is there anything I need to clean up before this is ready to go?

haf:
No, it's looking good. Was a bit confused since Adam took on reviewing. I'm merging and doing a release tomorrow

baronfel:
Excellent, thanks!

haf:
Released as v4.0.303!

PR #226, opened 2017-01-09
Feature/load tests

haf:
A bug is hiding in this code:

Registry.fs:

```fsharp

/// handling flyweights
module internal Logging =
  open Hopac.Infixes
  open System
  open Logary.Target
  open Logary.Internals
  open Logary.Internals.Date

  /// Flyweight logger impl - reconfigures to work when it is configured, and
  /// until then throws away all log lines.
  type Flyweight(name) =
    let state = IVar ()
    let instaPromise = Alt.always (Promise (()))
    let insta = Alt.always ()

    let logger (f : Logger -> _) defaul =
      if IVar.Now.isFull state then
        let _, logger = IVar.Now.get state
        f logger
      else
        defaul

    interface Named with
      member x.name = name

    interface FlyweightLogger with
      member x.configured lm =
        getLogger lm.registry name >>= fun logger ->
        state *<= (lm, logger)

    interface Logger with
      member x.logWithAck logLevel messageFactory =
        logger (fun logger -> logger.logWithAck logLevel messageFactory) instaPromise

      member x.log logLevel messageFactory =
        logger (fun logger -> logger.log logLevel messageFactory) insta

      member x.level =
        Verbose

  /// Singleton configuration entry point: call from the runLogary code.
  let startFlyweights inst : Job<unit> =
    // Iterate through all flywieghts and set the current LogManager for them
    Globals.withFlyweights <| fun flyweights ->
      flyweights
      |> List.traverseJobA (fun f -> f.configured inst)
      >>- fun (_ : unit list) ->
        Globals.singleton := Some inst
        Globals.clearFlywieghts ()

  /// Singleton configuration exit point: call from shutdownLogary code
  let shutdownFlyweights () =
    Globals.singleton := None
    Globals.clearFlywieghts ()
```

Logging.fs:

```fsharp

[<CompiledName "GetLoggerByPointName">]
let getLoggerByPointName name =
  if box name = null then nullArg "name"
  match !Globals.singleton with
  | None ->
    let logger = Flyweight name :> FlyweightLogger
    Globals.addFlyweight logger
    logger :> Logger

  | Some inst ->
    inst.runtimeInfo.logger.verbose (
      eventX "Getting logger by name {name}" >> setField "name" (name.ToString()))

    name
    |> Registry.getLogger inst.registry
    |> PromisedLogger.create name
```

And Globals.fs:

```fsharp

/// Module that is the ONLY module allowed to have global variables; created so
/// that consumer applications may call into Logary without having a reference
/// to the LogManager first.
///
/// The globals in this module are configured in module "Logging".
module internal Globals =
  open Logary

  /// This is the "Global Variable" containing the last configured
  /// Logary instance. If you configure more than one logary instance
  /// this will be replaced. It is internal so that noone
  /// changes it from the outside. Don't use this directly if you can
  /// avoid it, and instead take a c'tor dependency on LogaryRegistry
  /// or use IoC with a contextual lookup to resolve proper loggers.
  let singleton : LogaryInstance option ref = ref None

  /// This is the global console semaphore to use when printing in a multi-
  /// threaded manner to STDOUT or STDERR.
  let consoleSemaphore = obj ()

  /// A list of all loggers yet to be configured
  let private flyweights : FlyweightLogger list ref = ref []

  let private flyweightLock = obj ()

  let addFlyweight (fwl : FlyweightLogger) =
    lock flyweightLock <| fun _ ->
      flyweights := fwl :: !flyweights

  /// Gives f a snapshot of the current flyweights
  let withFlyweights f =
    f !flyweights

  let clearFlywieghts () =
    lock flyweightLock (fun _ -> flyweights := [])
```

With the tests in this PR.

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -20,12 +20,12 @@ NUGET
       elmah.corelibrary (>= 1.2.2)
       elmah.io.client (>= 2.0.29)
     EventStore.Client (3.9.4)
-    Expecto (3.0)
+    Expecto (3.0.1)
       Argu (>= 3.2)
       FSharp.Core (>= 3.1.2.5)
       Mono.Cecil (>= 0.9.6.4)
-    Expecto.FsCheck (3.0)
-      Expecto (>= 3.0)
+    Expecto.FsCheck (3.0.1)
+      Expecto (>= 3.0.1)
       FsCheck (>= 2.6.2)
       FSharp.Core (>= 3.1.2.5)
     FAKE (4.48)
@@ -116,8 +116,8 @@ GITHUB
     RingBuffer.fs (4557d607737e748b54eb6de3ec5e5cd02247b2a5)
       Hopac
   remote: logary/logary
-    src/Logary.CSharp.Facade/Facade.cs (397b5ee7733a36cb23c8887617470e427bbf1bb2)
-    src/Logary.Facade/Facade.fs (397b5ee7733a36cb23c8887617470e427bbf1bb2)
+    src/Logary.CSharp.Facade/Facade.cs (8fe3fce3f7d2602f2a4dc27419edfbce09c28a82)
+    src/Logary.Facade/Facade.fs (8fe3fce3f7d2602f2a4dc27419edfbce09c28a82)
 GROUP Examples
 REDIRECTS: ON
 FRAMEWORK: >= NET45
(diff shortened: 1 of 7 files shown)


lust4life:
https://github.com/logary/logary/blob/feature/live-wpc-reconfigure/src/tests/Logary.LoadTests/Logary.LoadTests.fs#L138-L144

@haf  try to migrate this test for the new version to see if it has the same problem, but didn't get the purpose of the test above, i write some comments there. can you help port them, or this bug is only bound to the old version's global flyweight implementation ?
```fsharp
// static get logger means always use the latest registry ('logary instance')
// so, we can not guarantee the output are in each registry's target output.
```

haf:
Yes, it's to see if we got memory leaks. I think we can drop it, but generally the software should be load-tested somehow, while tracking memory usage and when GC is reaping stale memory.

PR #232, opened 2017-01-29
[WIP] Facade: enable further customising output

### LINKED ISSUE #230: Facade LiterateConsole target: allow arbitrary log line prefix
In the Facade `LiterateConsole`, it currently only [supports a single way to render the log line prefix](https://github.com/logary/logary/blob/8fe3fce3f7d2602f2a4dc27419edfbce09c28a82/src/Logary.Facade/Facade.fs#L623) out to the console. However, in some cases, line prefix/template (e.g. `[10:01:55 INF] {message}`) doesn't make a lot of sense. It would be nice if could be easily configured differently.

@haf just raising this issue first in case you think it's not worthwhile. I obviously think it's worthwhile, and I don't *think* it will introduce too much complexity, but do you have any opinion at this early stage?

adamchester:
This will fix #230.

TODOs
 - [x] Add some tests that document how to customise the 'output template' as described in #230 
 - [ ] Verify that we want facade consumers to have the ability to parse event templates
 - [ ] Verify the breaking changes are acceptable (they are minor, e.g. `LiterateOptions` record fields changed).

### DIFF
--- a/src/Logary.Facade/Facade.fs
+++ b/src/Logary.Facade/Facade.fs
@@ -1,4 +1,4 @@
-/// The logging namespace, which contains the logging abstraction for this
+﻿/// The logging namespace, which contains the logging abstraction for this
 /// library. See https://github.com/logary/logary for details. This module is
 /// completely stand-alone in that it has no external references and its adapter
 /// in Logary has been well tested.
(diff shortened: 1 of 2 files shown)


adamchester:
@haf now would be a good time let me know what you think.

The basic idea is to keep everything the same, just allow more advanced customisation.

haf:
The thing is that we can't have breaking changes easily in the Facade without bumping a major version of all things that depend on the Facade, like Suave or Expecto, for example. Why do we have to have breaking changes?

One thing that stood out to me was the extensions module – which has to be opened manually – why?

adamchester:
@haf good points. I'll make another attempt at reworking this and pay closer attention to avoiding breaking changes.

Regarding the extensions, i need a way to keep the `create` method but also have the default `LiterateOptions` contain the implantation pieces.

haf:
Why can't the create method be in a module or in a type extension in a module that's auto-opened?

adamchester:
Ahh I see what you mean, yes it can and should be auto opened.

adamchester:
So we need binary compat here right? I think that means we can't modify record fields of`LiterateOptions` modify existing public signatures, right?

What about the `create` as an (auto-opened) extension method? Will that break binary compatibility?

haf:
Yes, but only for public interfaces. No, we can't modify public record fields or do breaking changes to public signatures.

Extension methods which don't conflict with existing methods are fine. So are adding functions to modules and changing the signature of internal or private things.

adamchester:
Closing this as I think it went too far down the wrong path. Hopefully next attempt will be better.

PR #234, opened 2017-02-04
Facade: allow customising the literate output template

adamchester:
This attempts to address #230

### DIFF
--- a/src/Logary.Facade/Facade.fs
+++ b/src/Logary.Facade/Facade.fs
@@ -1,4 +1,4 @@
-/// The logging namespace, which contains the logging abstraction for this
+﻿/// The logging namespace, which contains the logging abstraction for this
 /// library. See https://github.com/logary/logary for details. This module is
 /// completely stand-alone in that it has no external references and its adapter
 /// in Logary has been well tested.
@@ -685,7 +685,7 @@ module internal Formatting =
     formatFields matchedFields message.fields
 
 /// Assists with controlling the output of the `LiterateConsoleTarget`.
-module LiterateFormatting =
+module internal LiterateFormatting =
   open Literate
   type TokenisedPart = string * LiterateToken
   type LiterateTokeniser = LiterateOptions -> Message -> TokenisedPart list
@@ -699,7 +699,7 @@ module LiterateFormatting =
     tokens :> seq<TemplateToken>
   
   [<AutoOpen>]
-  module internal OutputTemplateTokenisers =
+  module OutputTemplateTokenisers =
 
     let tokeniseTimestamp format (options : LiterateOptions) (message : Message) = 
       let localDateTimeOffset = DateTimeOffset(message.utcTicks, TimeSpan.Zero).ToLocalTime()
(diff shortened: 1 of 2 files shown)


src/Logary.Facade/Facade.fs
```diff
@@ -684,6 +703,55 @@ module internal Formatting =
     formatExn message.fields +
     formatFields matchedFields message.fields
 
+/// Assists with implementing a custom tokeniser for the `LiterateConsoleTarget`.
+module LiterateFormatting =
```
haf: internal?

src/Logary.Facade/Facade.fs
```diff
@@ -684,6 +703,55 @@ module internal Formatting =
...
+
+  /// Creates a `LiterateTokeniser` function which can be provided to the `LiterateConsoleTarget`
+  /// constructor in order to render. The default template would be:
+  /// `[{timestampLocal:HH:mm:ss} {level}] {message}{newline}{exceptions}`.
+  /// Example of different output templates:
+  ///
+  /// `[{level}] {timestamp:u}: {message}
+  let tokeniserForOutputTemplate template : LiterateTokeniser =
+    let tokens = parseTemplate template
+    fun o m ->
```
haf: What's o an m?

src/Logary.Facade/Facade.fs
adamchester: Well, the extension point is in the `LiterateConsoleTarget` constructor already (`literateTokeniser`), but it's practically impossible to use it right now, because everything is internal. Making this public is intentional.

src/Logary.Facade/Facade.fs
```diff
@@ -515,6 +515,25 @@ module internal FsMtParser =
         go (ParserBits.findPropOrText start template foundTextF foundPropF)
     go 0
 
+module internal EventTemplateParser =
```
adamchester: I could probably delete this block now...

adamchester:
Pushed a few minor changes. What do you think of the general overall idea?

haf:
I like the overall idea very much. I understand what the intention is better now, too.

You should refactor `tokeniserForOutputTemplate` to use many small composable functions. Right now it's a big (clean) function but if I wanted to change the output printer I'd have to read through that code and re-do every bit I didn't want to change, like the printing of dates.

adamchester:
OK, I also gave this a try in the [Suave Example](https://github.com/SuaveIO/suave/tree/5cad201625d0179414f292abebfe5753dab52804/examples/Example):
```fsharp
let loggerTemplate = LiterateFormatting.tokeniserForOutputTemplate "[{level}] {timestampUtc:u} {message} [{source}]{exceptions}"
let singleLetterLogLevel = function Verbose->"V" |Debug->"D" |Info->"I" |Warn->"W" |Error->"E" |Fatal->"F"
let loggerOptions = { Literate.LiterateOptions.create() with getLogLevelText = singleLetterLogLevel }
let logger = LiterateConsoleTarget([|"Suave";"Example"|], Debug, options=loggerOptions, literateTokeniser=loggerTemplate) :> Logger 
```

It's looking good...

image

adamchester:
I think the `outputTemplate : string` should be a required parameter in a new `LiterateConsoleTarget` constructor overload. Then we can make the `LiterateFormatting` module `internal`.

```fsharp
  /// Creates the target with a custom output template. The default `outputTemplate`
  /// is `[{timestampLocal:HH:mm:ss} {level}] {message}{exceptions}`.
  /// Available template fields are: `timestamp`, `timestampUtc`, `level`, `source`,
  /// `newline`, `tab`, `message`, `exceptions`. Any misspelled or otheriwese invalid property
  /// names will be treated as `LiterateToken.MissingTemplateField`.
  new (name, minLevel, outputTemplate, ?options, ?outputWriter, ?consoleSemaphore) =
    let tokeniser = LiterateFormatting.tokeniserForOutputTemplate outputTemplate
    LiterateConsoleTarget(name, minLevel, ?options=options, literateTokeniser=tokeniser, ?outputWriter=outputWriter, ?consoleSemaphore=consoleSemaphore)
```

haf:
Ready to merge?

adamchester:
Ready!

haf:
Thank you @adamchester ! And thanks for testing it with Suave; that matters a lot.

PR #242, opened 2017-04-15
[facade] Don't render name when message has no name,

adamchester:
and fix up the tests to account for name rendering in the default literate output.


haf:
Thank you Adam!

Issue #133, opened 2016-07-04
Azure Application Insights Target


toburger:
Hi, I have a suave project running on Azure and a logary target for Azure Application Insights would be welcome.
I could write the target and make a PR, but ATM it has a low priority to us.
It should be similar to the AWS Cloud Watch target so I hope I can get some inspiration from that, but I'm pretty new to suave and also don't know much about Azure Application Insights, so please be patient with me. ;)

haf:
@toburger If you have any questions on how to build the target, don't hesitate to ask here.

haf:
I put up an offer here https://github.com/logary/logary#want-your-saas-logging-service-as-a-target if you're interested

christolo:
Any update on this? I'd be willing to help out if there's any interest.

toburger:
Sorry for not replying on this topic anymore.
Due to time constraints on our project we've chosen to use Isaac Abrahams solution which turns out to work very well for us.
Here is an example of the solution: https://github.com/isaacabraham/azure-fsharp-helpers

christolo:
Great, thanks! I'll check it out.

Thorium:
I'm also interested of this... But I'm not using Suave. 
Maybe this would be just a few lines of code to write a target. 
Something like https://github.com/Microsoft/ApplicationInsights-dotnet-logging
I'll dig this a bit more soon.

PR #244, opened 2017-04-18
Microsoft Azure Application Insights target for Logary.

Thorium:
See all the Logary logged messages in Microsoft Application Insights TRACE-events.
Allows more details to use-case #133
For usage example, see `src\targets\Logary.Targets.AppInsights\Script.fsx`

Working:
sample

How should this be released?
@haf are you the one to release targets?
Feel free to add to Nuget if so.
Do you create all the packages as separate Nuget package?


haf:
Yes, our build script automatically creates packages from the DLLs. Currently the build is failing on Linux, probably due to the FS being case-sensitive, with:

```
Paket failed with:
	The file "src/targets/Logary.Targets.AppInsights/bin/Release/Logary.Targets.AppInsights.xml" requested in /home/travis/build/logary/logary/paket.template does not exist.
rake aborted!
Albacore::CommandFailedError: Command failed with status (1):
  mono tools/paket.exe pack output build/pkg
```

Do you accept to maintain this target, if I add it to the code-base? Maintenance primarily revolves around bumping its dependencies once in a while and porting it when new major version of Logary are released (every other year approximately).

Thorium:
I can maintain.

haf:
Thank you!

Thorium:
And howabout NuGet package release? Do I have to do still something now?

haf:
Tomorrow I'll release it :)

PR #247, opened 2017-04-23
Facade/Literate output properties

adamchester:
Adds support for `{properties}` and `{newLineIfNext}` in the `LiterateConsole` `outputTemplate`. Currently, if libraries like Suave write a log `Message` out with any **fields that are not included in the message template**, they will not be rendered and there is no way to easily enable it. 

This feature would allow you add `{properties}` to the output template:
```fsharp
let outputTemplate =
  "[{level}] {timestampUtc:u} {message} [{source}]{newLineIfNext}{properties}{exceptions}"

let singleLetterLogLevel = function
  | Verbose-> "V"
  | Debug-> "D"
  | Info-> "I"
  | Warn-> "W"
  | Error-> "E"
  | Fatal-> "F"

let loggerOptions =
  { Literate.LiterateOptions.create() with
      getLogLevelText = singleLetterLogLevel }

let logger = LiterateConsoleTarget(
              [|"Suave";"Example"|],
              Verbose,
              options=loggerOptions,
              outputTemplate=outputTemplate) :> Logger 
```

The following screen shot shows the line ` - cookieName: counter`; it is an example of an important field that is part of the log `Message` but **not part of the message template**. Without this feature, we cannot see the `cookieName`.

image

~(Note: I'm still experimenting with this idea a little. I'm raising PR now to get high level feedback).~

### DIFF
--- a/src/Logary.Facade/Facade.fs
+++ b/src/Logary.Facade/Facade.fs
@@ -299,14 +299,6 @@ type LoggingConfig =
     /// adapter to replace this semaphore with a global semaphore.
     consoleSemaphore: obj }
 
-[<AutoOpen>]
-module internal Helpers = 
-  /// replaces the built-in `isNull` function so we can still use it when < FSharp.Core 4.0
-  let inline isNull o =
-    match o with
-    | null -> true
-    | _ -> false
-
 module Literate =
   /// The output tokens, which can be potentially coloured.
   type LiterateToken =
--- a/src/Logary.Facade/Facade.fs
+++ b/src/Logary.Facade/Facade.fs
@@ -300,6 +300,14 @@ type LoggingConfig =
     /// adapter to replace this semaphore with a global semaphore.
     consoleSemaphore: obj }
 
+[<AutoOpen>]
+module internal Helpers = 
+  /// replaces the built-in `isNull` function so we can still use it when < FSharp.Core 4.0
+  let inline isNull o =
+    match o with
+    | null -> true
+    | _ -> false
+
 module Literate =
   /// The output tokens, which can be potentially coloured.
   type LiterateToken =
@@ -381,11 +389,6 @@ module internal FsMtParser =
     override x.ToString() = x.AppendPropertyString(StringBuilder()).ToString()
 
   module internal ParserBits =
-
-    let inline isNull o =
-      match o with
-      | null -> true
-      | _ -> false
 
     let inline isLetterOrDigit c = System.Char.IsLetterOrDigit c
     let inline isValidInPropName c = c = '_' || System.Char.IsLetterOrDigit c
--- a/src/Logary.Facade/Facade.fs
+++ b/src/Logary.Facade/Facade.fs
@@ -8,7 +8,6 @@
 namespace Logary.Facade
 
 open System
-open System.Runtime.CompilerServices
 
 /// The log level denotes how 'important' the gauge or event message is.
 [<CustomEquality; CustomComparison>]
@@ -75,7 +74,7 @@ type LogLevel =
     | 4 -> Warn
     | 5 -> Error
     | 6 -> Fatal
-    | _ as i -> failwithf "LogLevel matching integer %i is not available" i) i
+    | i -> failwithf "LogLevel matching integer %i is not available" i) i
 
   interface IComparable<LogLevel> with
     member x.CompareTo other =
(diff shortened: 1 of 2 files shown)


haf:
A small merge is needed. I'd love to merge this PR.

adamchester:
Great, I'll push forward soon...

haf:
Is this ready to merge?

adamchester:
Thanks for pinging me on this one; actually even though I've left this lingering for quite a while now, I do want to finish it off. I'm aiming to fix it up within a couple of weeks.

btrepp:
This would be great for dropping timers everywhere, Eg when using Message.timeAlt etc. I usually use this and add properties to the message (eg the Id of the object etc), This becomes handy in influx, but A bit obtuse in the console eg.

PointName took x ms to execute.

lust4life:
image

source:  https://github.com/logary/logary/blob/feature/live-wpc-reconfigure/examples/Logary.Suave/Program.fs

@adamchester Can you consider do some more optimization improvements based on this branch ?

haf:
Ping @adamchester

haf:
@adamchester Could you please rebase?

adamchester:
ok, attempting to rebase this now 🎉 wish me luck!

adamchester:
Update: It's not going well so far. I'm currently struggling to build `master` on windows. I think it's due to `.paket\paket.exe` being old and it's unable to connect to github due to some security errors. 

```shell
C:\My\repo\logary-2>build
The request was aborted: Could not create SSL/TLS secure channel. (Github - cached (temporarily ignore updates))
The system cannot find the path specified.
```

Will try updating paket now.

adamchester:
Hey @lust4life, big delay in replying to you here sorry! It wasn't clear to me what [you meant here above](https://github.com/logary/logary/pull/247#issuecomment-350810241). Any chance you could elaborate a bit?

haf:
The current build should compile but I'm working through the unit tests; of which most succeed.

adamchester:
Thanks @haf, it looks like I have a compilation error introduced during the rebase to fix first too.

lust4life:
> Hey @lust4life, big delay in replying to you here sorry! It wasn't clear to me what you meant here above. Any chance you could elaborate a bit?

see here https://github.com/logary/logary/blob/master/examples/Logary.Suave/Program.fs ,

i mean right now logary rewrite its formatting (Json, MessageTemplates based your `FsMtParserFull`) can output properties, and at that time the code have change a lot, i was worried you will continue with the old version (v4), so i say this: 
> Can you consider do some more optimization improvements based on this branch`

but this branch has been merged into master, you can directly based on the master branch now.

adamchester:
This could use a quick review now, or we can just push the button ...?

src/Logary.Facade/Facade.fs
```diff
@@ -297,6 +297,14 @@ type LoggingConfig =
     /// adapter to replace this semaphore with a global semaphore.
     consoleSemaphore: obj }
 
+[<AutoOpen>]
+module internal Helpers =
```
haf: I don't think this should go here; it's not for Logary to provide backwards-compatibilities; if the author using the facade wants this he'll import a shim of his own and place it ahead of the facade. Plus, you don't capture the full extent of isNull **I think**, or else the recommended way would be `if o = null` by the same logic you have in your implementation.

haf:
Quick fix; remove `isNull`, then we merge!

src/Logary.Facade/Facade.fs
adamchester: No probs, I’m only moving it here, I didn’t add it. I can remove it a bit later when I’m back at the computer, but it came from here originally https://github.com/logary/logary/commit/fb7f4398f4895a7c02483d08163ac1aeee737a66

haf:
Yes I saw that; missed it in the large previous changesets, but I think we should remove it.

haf:
Thank you @adamchester !

haf:
@adamchester Is this available in Logary proper as well?

adamchester:
Hey @haf, not exactly the same, no. This PR deals with the `outputTemplate` and `{properties}`.

However, in Logary proper, there is already something very similar; one could compose a new `tokenise` function from all the smaller pieces, those tokenise functions are very similar to what is here (but probably more powerful).

haf:
@adamchester How would I write a tokenise function that doesn't output newlines?

PR #253, opened 2017-05-12
make c# api conf more flexible when config internal log

lust4life:
api config for c# is only support console target for logging internal logs right now , can add flexible for the case use asp.net in vs, can use debugger or some other targets for internal logging.

```csharp
LogaryFactory.New("demoService",
                    conf => conf
                            .InternalLoggingLevel(LogLevel.Verbose)                        
                            .Target<Debugger.Builder>("debugger", tb => tb.UseForInternalLog())
                            .Target<Logary.Targets.Console.Builder>("console", tb => tb.UseForInternalLog())
                            .Target<LiterateConsole.Builder>("console1")                            
                            .Target<AliYun.Builder>("AliYunLog", BuildAliYunTarget)
                    );
```

- [x] add sample code in readme

- [ ] add some test

```fsharp
    // generate msg use default messagetemplates
    let formatedMsg = Formatting.format (Parser.parse formatTemplate) args
```
csharp LoggerExtensions when log event with object args ,  when use fsharp , there is no chance to add othere type in `fields    : Map<PointName, Field>` ， but when use in csharp  with the `LogEventFormat` method , we use `args : obj[]` for templete . 

use `MessageTemplate` as default format, when invoke `LogEventFormat` from c# api.

### DIFF
--- a/src/Logary/Configuration_FactoryApi.fs
+++ b/src/Logary/Configuration_FactoryApi.fs
@@ -20,11 +20,15 @@ open Logary.Configuration
 open Logary.Target
 open Logary.Targets
 open Logary.Target.FactoryApi
+open System.Text.RegularExpressions
+
 
 type internal ConfBuilderT<'T when 'T :> SpecificTargetConf> =
-  { parent      : ConfBuilder        // logary that is being configured
-    tr          : Rule               // for this specific target
-    tcSpecific  : 'T option }
+  { parent            : ConfBuilder        // logary that is being configured
+    tr                : Rule               // for this specific target
+    tcSpecific        : 'T option
+    useForInternalLog : bool               // make internallog target config flexible
+  }
 with
   member internal x.SetTcSpecific tcs =
     { x with tcSpecific = Some tcs }
@@ -37,6 +41,10 @@ with
 
     member x.SourceMatching regex =
       { x with tr = { x.tr with Rule.hiera = regex } }
+      :> TargetConfBuild<'T>
+
+    member x.UseForInternalLog () =
+      { x with useForInternalLog = true }
       :> TargetConfBuild<'T>
 
     member x.AcceptIf acceptor =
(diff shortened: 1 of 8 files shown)


src/Logary/Configuration_FactoryApi.fs
```diff
@@ -57,14 +65,51 @@ and internal MetricsConfBuilder(conf) =
...
+  let configInternalTargets conf =
+    // find all targets set for internal log through a specific rule,
+    // and remove these (specific rule/target) from logary conf, so normal logging is not affected. 
+    // if there isn't any, use console as default
+
+    let (normalRules, forInternalRules) =
+      conf.rules 
+      |> List.partition (fun r -> 
+            match r with
+            | r when r.hiera.ToString() = specificHieraForSetInternalLog -> false
```
haf: A hiera is a regex and should be used as such.

src/Logary/Configuration_FactoryApi.fs
```diff
@@ -57,14 +65,51 @@ and internal MetricsConfBuilder(conf) =
 
 /// The "main" fluent-config-api type with extension method for configuring
 /// Logary rules as well as configuring specific targets.
-and ConfBuilder(conf) =
+and ConfBuilder(conf, internalLoggingLevel) =
+
+  let specificHieraForSetInternalLog = "^specificHieraForSetInternalLog$"
```
haf: Why can't we represent this case in types?

src/Logary/Configuration_FactoryApi.fs
```diff
@@ -162,5 +214,5 @@ type LogaryFactory =
     if serviceName = null then nullArg "serviceName"
     if configurator = null then nullArg "configurator"
     let config = Config.confLogary serviceName
-    let cb = configurator.Invoke (ConfBuilder config)
-    cb.BuildLogary () |> CSharp.toTask
\ No newline at end of file
+    let cb = configurator.Invoke (ConfBuilder (config,LogLevel.Fatal) )
```
haf: Where's this level used?

src/Logary/Configuration_FactoryApi.fs
```diff
@@ -128,10 +175,15 @@ and ConfBuilder(conf) =
     // referentially transparent
     let targetConf = f.Invoke(!contRef) :?> ConfBuilderT<'T>
 
+    let tr = 
+      match targetConf.useForInternalLog with
```
haf: Prefer if-s

src/Logary/Configuration_FactoryApi.fs
```diff
@@ -57,14 +65,51 @@ and internal MetricsConfBuilder(conf) =
...
+          forInternalRules 
+          |> List.exists (fun r -> r.target = k.ToString())
+          |> not)
+
+    let internalTargets = 
+      internalTargets 
+      |> Map.toList
+      |> List.map (fun (_,tci) -> fst tci) 
+      |> function 
+        | [] -> [Console.create Console.empty "internalConsole"]
```
haf: Can't we default to this in the building?

lust4life: it is the default , see comments after.

src/Logary/Configuration_FactoryApi.fs
lust4life: it is the default internal logging level , use when people not setting it from `InternalLoggingLevel` . see comments after.

src/Logary/Configuration_FactoryApi.fs
lust4life: here is a trick, in order to identity which targets are set for internal logging.  the specific rule with hiera `specificHieraForSetInternalLog ` is for matching a target through its `rule.target` property. 

not find other good way to do it , maybe we can use some conventions to do it , like:  a target with name match some  hiera is the internal target . but maybe affect some exists code ?

src/Logary/Configuration_FactoryApi.fs
lust4life: more details ? did not get it ☹

src/Logary/Configuration_FactoryApi.fs
haf: Seems like a boolean

haf:
+1

src/Logary/Configuration_FactoryApi.fs
lust4life: done.

haf:
WIP or ready?

lust4life:
ready.

haf:
Thank you @lust4life !

Issue #252, opened 2017-05-11
source code question about registry


lust4life:
https://github.com/logary/logary/blob/master/src/Logary/Registry.fs#L106-L122

```fsharp
    // rules applying to the same target are grouped
    |> Seq.groupBy (fun (r, t, ti) -> t.name)

    // combine acceptors with Seq.any/combineAccept
    |> Seq.map (fun (_, ts) ->
        let _, t, ti = Seq.head ts
        let rs       = Seq.map rules ts
        let minLvl   = rs |> Seq.map (fun r -> r.level) |> Seq.min
        // find the min matching level from all rules for this target
        createFilter minLvl rs,
        t, ti,
        minLvl)

    // targets should be distinctly returned (deduplicated, so that doubly matching
    // rules don't duply log)
    |> Seq.distinctBy (fun (_, t, _, _) ->
        t.name)

```

seems `distinctBy` is not necessary beacuse of groupBy target name ?

```fsharp
  let inline setContext name value message =
    Lens.setPartial (Lenses.contextValue_ name) (Value.serialize value) message
```

when invoke it from c#, if value is a POCO , it will throw a `NotSupportedException` , maybe we can add a `ToValue` for `POCO` in ToValueDefaults ? maybe use `Value.ofObject` or just `ToString` ?

lust4life:
@haf  ping.

haf:
I think the code *is* necessary, because I test-drove it and I need to write it. You can see if the tests pass if you remove it. I don't pretend I can keep that code fully executing under all circumstances in my head. ;)

You mean overloading ToValue for `object`? Then you can't use the other overloads.

In the new-engine branch; I'd love help with adding a 'known types' registry that is configurable per Logary instance; all others should have ToString() called on them.

lust4life:
> You mean overloading ToValue for object? Then you can't use the other overloads.

get it tks.  seems `value.ofObject` handles all the other overloads but need do more  pattern match the `value.ofObject` result and `set value LensPartial` will cover it , one big `ToValue` method to rule them all. looks ugly.  😄

lust4life:
try tests later, because some ruby issue on windows,  can't run `bunlde exec rake` easily .

PR #254, opened 2017-05-12
test for remove distinctBy beacuse of groupBy

lust4life:
- and add test for `doubly matching rules or mutil targets with same name should not log deduplicated`


lust4life:
should close #252

Issue #246, opened 2017-04-21
Application Insights Targets feedback


SergeyKanzhelev:
Thank you for the targets for Application Insights. Here is some feedback:

- [x] ~~I'd suggest to rename `AppInsights` to `ApplicationInsights` in NuGet name to avoid confusion. AppInsights is a different tool~~
- [x] ~~Also suggest to use full name for `AppInsightsKey`~~
- [x] ~~typically logger should take dependency on base SDK (`Microsoft.ApplicationInsights`). Reference to `Microsoft.ApplicationInsights.WindowsServer` will be added by user.~~
- [ ]  If you want to keep dependencies collection - please also add a standard telemetry initializer `HttpDependenciesParsingTelemetryInitializer` (like https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/401)
- [ ] Is it typical to target `4.5.2` for F# projects? Why not `4.6` or `4.5`?
- [x] ~~give a link to https://docs.microsoft.com/azure/application-insights/app-insights-create-new-resource in error message and comment `Enter your instrumentation key here. Get it from Azure Portal -> App Insights -> Properties -> INSTRUMENTATION KEY`~~
- [x] ~~Set `.Context.GetInternalContext().SdkVerion = "logary: <logary sem-version>"` to indicate that traces were collected by this adaptor. It will simplify troubleshooting if something goes not like expected.~~

haf:
@Thorium I've given you write access to the repository to manage this target. :)

Thorium:
Ok, thanks!

haf:
<q> Is it typical to target 4.5.2 for F# projects? Why not 4.6 or 4.5?</q> – Not really – does it make a difference?

SergeyKanzhelev:
@haf, target 4.5.2 force people to upgrade projects from 4.5, but do not go all the way to 4.6 where some new features like better correlation support are available.

SergeyKanzhelev:
@Thorium I'll be happy to help with Application Insights questions. Ask in https://github.com/Microsoft/ApplicationInsights-Home/issues or mention me in issues.

Thorium:
Is there more info about HttpDependenciesParsingTelemetryInitializer?

Logary targets were in general 4.5.2 so that's why I used it. I think it's convenient: It doesn't force you to do full update to .NET 4.6 (which can be problematic with some earlier MONO versions) and still it has many fixes that there has been for the old 4.5 for some time already.

[...]

PR #261, opened 2017-07-25
Better format value without unit in Application Insights Target

xuanduc987:
@Thorium 
As commented [here](https://github.com/logary/logary/issues/246#issuecomment-316907468), this PR uses `Formatting.MessageParts.formatValue` instead of `Units.formatValue` for field without unit.


Thorium:
Thanks!

Issue #257, opened 2017-05-14
event format proposal


lust4life:
>In the new-engine branch; I'd love help with adding a 'known types' registry that is configurable per Logary instance; all others should have ToString() called on them.

#252   mentioned here, 

```f#
type Field =
  Field of Value * Units option 
```
i have a question, why Field have this signature , `Value * Units option`,  can we just use  `obj * Units option` , and if we have this signature, we can use FsMessageTemplates default format method ,and have the same consistency about (`$,@,align...`), and `adding a 'known types' registry` can by done by pass textwriter with `System.ICustomFormatter` ? 

use obj can have benefit when people what to custom their own `ToString` method， but this may can be done by `adding a 'known types' registry `  ?

another problem is , when we use `Value` ， we need call `Value.ofObject` ，it need support **ReferenceLoop** handling ? maybe ignore referenceloop or use a default maxDepth like **MessageTemplates ** , then it will not throw `StackOverflowException` when the object has a circle reference.

haf:
Right; and the obj could still be of value `Value`, but boxed.

Then add the step/registry that converts to type `Value` from `obj`, doing ref-counting.

Then feed it into the targets.

Good.

lust4life:
```fsharp
/// Logary's structured logging is centered around the `Value` type.
///
/// All objects logged into Logary are translated into a structure such as this
/// one.
///
/// The trade-off between passing all the CLR type information along versus
/// fitting it all into these cases is one of completeness versus convenience
/// in writing targets and computing on the data.
type Value =
  | String of string
  | Bool of bool
  | Float of float
...
```


in the new branch, i see some comments about `Value` type,`convenience` is good for `Gauge/Derived` doing some computing about metrics. but for `Event`, it has some limits, and the `fields` prop in `Message` make them loose coupling, `fields` should only for `Event` template rendering.

so shall we make some change like this:

```fsharp
type PointValue =
  /// Value at point in time
  | Gauge of Value * Units
  /// Any sort of derived measure
  | Derived of Value * Units
  /// All simple-valued fields' values can be templated into the template string
  /// when outputting the value in the target.
  | Event of template:string * args: obj []
```

the `args` is only for template rendering, we can supply an default format implementation use `FsMessageTemplates.sprintsm` (signature: `System.IFormatProvider -> string -> obj[] -> string`).

and every `Target` has its own `IFormatProvider` for this implementation. clients can add their own custom type render registry. maybe like `TemplateRender` below :

```fsharp
open System
[...]

PR #263, opened 2017-08-04
WIP: Feature/live wpc reconfigure

lust4life:
- remove pollServices in registry

  Because supervisor will supervise minions, if an exception occurs, it will be handled according to policy. So there's no need for state polling, because we can't do much if we can get the status. the user's handle logic can be expressed by policy. 

- remove healthcheck & metric

  >  The aim here is to normalise the Gauge/Derived/Message semantics into a single semantic of 'things computed from events from your app'.

  - remove healthcheck 

    Looks like metric can also be seen as a healthcheck, such as windows performance counters. when outside tick , metric generate some msg, which will be processed by user defined engine , they can decide their meaning (log or metric or healthcheck...) there.

  - remove metric 
    
    metric can be implemented by engine's process expression builder, what we need to do is schedule jobs for tick, in the corresponding time(like old metricConf.tickInterval) or manually tick. Metric's data sources can have two types, one is similar to WPC, generate data in real-time when ticked .the other is generate through the normal log message.

- support multi targets on internallogger and apply rules when send msg to target
- remove loglevel on Logger interface

  message has the log level info, targets have the rules info, before putting msg to target, apply rules on message to decide if really send to target.

- todo 

  - [x] ExpWeightedMovAvg -> events ticker
  - [ ] more unit test
     - [ ] ewma & percentile
     - [x] use pipe builder build an healthchecker
  - [x] support message filter, processing reconfiguration
         can be achieved by create another logary instance, but needs user create logger through static create, if user create log through an logary instance, things does not work for this situation. and we should think about how to stop ticker jobs on pipes, see below.
  - [x] targets rule to event processing
    - [ ] if discussion blow passed, need remove rules on target
  - [ ] event format improve mentioned by #257 
  - [ ] complete todo comments part in codebase
  - [ ] **think about how to GC ticker jobs on events, these may leak mem** ?


lust4life:
@haf  add some temporary code `Events.fs` for discussion. 

> This execution/filtering/windowing, is what I want build in a synchronous manner with Flow, we either throw the message away or we drop it into a RingBuffer.

really don't know how to ` build in a synchronous manner with the current Flow` , beacuse flow need a source, but what we really want is just function composition ？ flow can only pass a context and run iterator ( tryAdvance or bulk), witch just iterator the source elements.  how can we define the source elements ?

and is logary's own targets dispatching need to implement in `Processing` ?

src/Logary.CSharp.Facade/Facade.cs
```diff
@@ -471,15 +471,16 @@ public Flyweight(string[] name)
...
                 {
                     lock (_updating)
                     {
-                        _logger = local.Item1.GetLogger(_name);
-                        _fwClock += 1u;
+                        var local = _config;
+                        if(_fwClock != local.Item2)
+                        {
+                            _logger = local.Item1.GetLogger(_name);
+                            _fwClock = local.Item2;
```
haf: Why not bump the clock?

src/Logary.Facade/Facade.fs
```diff
@@ -783,7 +783,7 @@ module Global =
           let cfg, cfgClock = !config
           if cfgClock <> fwClock then
             logger <- cfg.getLogger name
-            fwClock <- fwClock + 1u
+            fwClock <- cfgClock
```
haf: Same here... Why?

src/Logary/Internals/Global.fs
```diff
@@ -75,7 +75,7 @@ module internal Global =
           let cfg, cfgClock = !config // reread the config's clock after taking lock
           if cfgClock <> fwClock then // recheck after taking lock to avoid races
             logger <- cfg.getLogger name // get the current logger
-            fwClock <- fwClock + 1u // update instance's clock
+            fwClock <- cfgClock // update instance's clock
```
haf: ...and here...

src/Logary/Registry.fs
```diff
@@ -62,29 +62,51 @@ module Engine =
...
+          processing (messageFactory level) emitCh
+          ^=>. loop subsribers
 
         emitCh ^=> fun message ->
-          inputCh *<- (message.level, fun _ -> message)
+          // send to targets
+          let targetName = Message.tryGetContext "target" message
+          match targetName with 
+          | Some (String targetName) ->
+              let subscriber = HashMap.tryFind targetName subsribers
```
haf: 2 space indentation

src/Logary/Registry.fs
```diff
@@ -62,29 +62,51 @@ module Engine =
...
+      subscriberCh : Ch<string * (Message->Job<unit>)> 
     }
 
   let create (Processing processing) : Job<T> =
-    let inputCh, emitCh, shutdownCh = Ch (), Ch (), Ch ()
-    let callback = Ch ()
-    let rec loop () =
+    let inputCh, emitCh, shutdownCh, subscriberCh = Ch (), Ch (), Ch (), Ch ()
+
+    let engine = { subscriptions = HashMap.empty
```
haf: Format on new line

haf:
I think the PR is going in the right direction, and I understand it's primarily about discussion right now.

>  beacuse flow need a source, but what we really want is just function composition

Yes; I was thinking an infinite source that's a synchronous/isolated/threaded thing, taking messages from the surrounding system. The "processing" function is then composed with ` (source |> flatMap processing)` or something similar.

```
module Engine =
  let create () =
    // bridge async/sync gap
    let rec source =
      Job.server (Job.isolate (job {
      while hasValues buffer do
        Interlocked.Increment(&chaser, 1)
      }))
      // in the engine loop:
      Flow.run (readBuffer chaser head buffer)
      // then somewhere below
      Job.result engine

let engine =
  memo (Engine.create ())

let impl level fac: Alt<Promise<unit>> =
  engine >>= fun engine ->
  Alt.prepareFun <| fun nack ->
  // allocate index
  let i = ref 0
  let p = Promise ()
  while Interlocked.CompareAndSwap(&head, ...) <> ... && not (Promise.isFulfilled nack) then
    buffer.[i] <- (level, fac, p)
  p

let logger = { new Logger with member x.logWithAck level fac = impl level fac }

let usage =
  let a = fn ()
  logger.info (eventX "Thing {a}" >> setField "a" a)
```

Then `Flow.run` would block on the source; thereby dedicating a single thread to the single processing step in Logary.

Targets implement processing? No, I'm thinking something like (pseudo code):

let processing =
  Flow.map enrich
  >> Flow.groupBy (fun x -> x.fields.['service'])
  >> Flow.window 5s
  >> Flow.sink (target "elasticsearch")
```

Each sink could then be a source of a de-multiplexer/fan-in function that writes to the RingBuffers of the named targets (found by config lookup). `Flow.sink` should block synchronously if the buffers it is writing to are full and the policies for them aren't allowing dropping messages.

File: src/Logary.CSharp.Facade/Facade.cs (hunk 1 above)
Comment (lust4life) [inline src/Logary.CSharp.Facade/Facade.cs]: every time we initialize config, the clock increases by 1. and when we log message, we will check if the local config clock same as the current config's clock , if not, get new logger from new config and update the clock then log message use the new logger .

e.g.
1. the default fwClock is 1.
1. initialize logging config ,  config clock will be 2
1. if we initialize logging config twice here or between some time,  config clock will be 3
1. then log message through facade api, it should be updating config and clock once. 

so assign the current clock to the fwClock (not by adding one) can avoid more checking .

File: src/Logary.Facade/Facade.fs (hunk 2 above)
Comment (lust4life) [inline src/Logary.Facade/Facade.fs]: see reply above

**Comment by lust4life** [comment]
@haf  

 are we going  build an `processing` through `Flow/Events expression` to achieve something like `Logary.Transformers` (`Counters,Reservoirs,ExpWeightedMovAvg,SlidingWindow...`) ?


previous question :
> and is logary's own targets dispatching need to implement in Processing ?

i mean in previous version , when we get a logger, we will find the corresponding targets, and  put message to targets buffer when logging. but now, when we get a logger and logging, all message will go to the `engine` and through `processing` , and put to targets buffer if the message has `sink` to a specific target . what If the user does not provide any processing, will all the messages be discarded ?

**Comment by haf** [comment]
Yes, then all messages are discarded; because the new processing will take the place of Rule and the configuration of where messages go.

**Comment by lust4life** [comment]
thanks for confirming, it is getting clearer. will try to carry on

**Comment by lust4life** [comment]
@haf  Can you help me check if `Pipe` in `Events.fs` meets the requirements?

**Comment by lust4life** [comment]
@haf  

I was thinking, when we implemented the event processing builder, the logger itself became stateless, just an `logging API` feed message into engine. If so, there is no need to provide `getlogger` interfaces on `LogManager` . we can make `Global.getStaticLogger`  lighter. no need to communicate through `client-server (promise wrappers)`, what we need for logger is  `middlewares in logaryConf` and the current 'engine'  . so we can capture them by a function when we initialize `Registry` , inject it into `Global.getLogger` .

if that, message middleware compose and messagefactory is invoked at user's call-site. issue like this #255  can be solved.

**Comment by haf** [comment]
> the logger becomes stateless

It doesn't have any state, but a reference to the engine.

> just the API

yes

> getLogger on LogManager

Is needed, because you may have multiple logary instances running concurrently and the LogManager/Registry/Engine needs an API to expose to the user.

> getStaticLogger lighter

How? Given the above requirement for not-just-globals.

> No need to communicate client-server

But there still is; we still need a way of accepting the message into a given target and then ack on send.

**Comment by lust4life** [comment]
change my previous thoughts a bit , yes you are right, the api is need, because multiple logary/registry instances. 
but no need to be an async(fake sync) style.  

commit some changes, if you have time to look at it .

**Comment by haf** [comment]
We need to provide the ability to wait for an ACK for a specific target (this is what the changes to the target configuration/policies were about). That's async. But the initial treatment can be avoided; that said, what happens when Logary fills up due to a slow target?

**Comment by lust4life** [comment]
> We need to provide the ability to wait for an ACK for a specific target
i think get/create a logger is something about expose logging api though an engine. 

if create request is called by static (e.g. `Logging.getCurrentLogger`) then it goes to the current engine capture by `Global.config`. when initialize another logary instance, engine update handled by `Flyweight`.

if create request is called by each logmanager instances running concurrently, then it goes to their own engine.

This process does not require target intervention.

> what happens when Logary fills up due to a slow target
the `log/logack`  is async, these logging call on a logger handle this situation. when logging, 
- messageFactory generate msg at call-site first, then feed it into engine wait for ack.
- engine process msg through processing logic predefined by event processing builder.
- finally, msg may(may not) go to the target buffer waiting for target send ack

nothing changes here.

**Comment by haf** [comment]
@lust4life Then we agree, good.

**Comment by lust4life** [comment]
i think the `registry` part is almost done yet. and the `event processing builder` part is basically done.

i will try to write some test for showing its usage, if you think it can meet our requirements on this stage, i will continue port the other parts (metrics,targets, or other projects broken by this pr)

after that, we can improve event format mentioned before #257

**Comment by haf** [comment]
Absolutely; I'm a bit tied down right now in https://github.com/fsprojects/Paket/issues/2616#event-1210825661 which has basically my ability to do productive work right now. But when this clears up and I've worked through the backlog created, I'll try this out.

**Comment by lust4life** [comment]
- [ ] targets rule to event processing

i am wondering do we need keep rules in targets, support an compatible `ConfigFactoryApi` for c# ? in there when config a target, we can set a rule on it to decide which message this target need concern.

but now, we support processing, instead of having to support rules on target, let the user filter the message directly through the event processing builder. i think this way is more natural, because target itself does not need to know which message they should process. user should define these rules somewhere else (`processing builder`).

**Comment by haf** [comment]
I would like rules **on targets** but not for the routing thereto. Rules such as what drop policity to use and and retry/respawn policies to use. These rules are primarily centred around *config* and not *message routing*, so it's probably alright to drop the rules, if we have an equivalently expressive processing "config"

**Comment by haf** [comment]
As for the C# stuff; let's figure that after figuring out a nice processing abstraction :)

**Comment by lust4life** [comment]
@haf  add some test for Event Processing : [Tests/Engine.fs](https://github.com/lust4life/logary/blob/0b7d53e97980b42078843b034502549252a7f501/src/tests/Logary.Tests/Engine.fs) 

because there are some `pipe/events processing` involving `time, rate`, it is difficult to accurately test. So the test method looks a little ugly. but i think it can show the main idea.

something like this: 
```fsharp
       let processing = 
          Events.stream 
          |> Events.subscribers [
            Pipe.start |> Events.service "svc1" |> Events.counter (TimeSpan.FromMilliseconds 100.) |> Events.sink "1"

            Pipe.start |> Events.tag "gotoTarget2" |> Pipe.map (fun msg -> { msg with value = Event ":)"} ) 
            |> Events.sink "2"

            Pipe.start |> Events.miniLevel Warn |> Events.sink "3"

            Pipe.start |> Pipe.bufferTime (TimeSpan.FromMilliseconds 200.) 
            |> Pipe.map (fun msgs -> Message.event Info (sprintf "there are %i msgs on every 200 milliseconds" (Seq.length msgs)))
            |> Events.sink "4"
          ]
       |> Events.toProcessing
```

haf:
We should have the `IClock` interface extracted everywhere so that time can be ticked manually. My environment is improving and I can almost always use F# now.

lust4life:
right now, `time auto tick` ticker is implement by `hopac.timeout` which rely on `Environment.TickCount` , but ticker itself can support manually tick, so seems in theory we can simulate `time auto tick` for just test. 

but i think `processing` here is not intended for high precision timing, i just print info to console to see whether the effect is roughly right or not 😅.

> My environment is improving and I can almost always use F# now

🎉

lust4life:
@haf i was thinking,  `SamplePeriod` in `ExpWeightedMovAvg` now is 5 seconds, and is used for calculate `alpha` and `instantRate`, should we calculate them (`alpha`,`instantRate`) real-time based on every tick's real interval?

lust4life:
@haf  [add percentile and EWMA event processing](https://github.com/logary/logary/pull/263/commits/1e1efe02ca5685294651481fb67d2de1b8e0dc0c) ,

> Metrics/Streams
> 
> The aim is to port something of Nessos Streams into a metrics engine, so that the configuration of where to send what names/hieras/messages/metrics/etc is done with code rather than the rules. That way we can add in calculating metrics, like finding the 99th percentile of request latency every rolling 5 second window.

i think aim here can work now.

lust4life:
Ping ?

haf:
@lust4life Yes, that could work. How do would you design the filtering and end-to-end declaration of the processing pipeline?

I don't really understand the code though; you have `fiveMinutesEWMATicker.TickEvery (Ti...` but then you have another line with `tick`? Why do we have an instance of a ticker used twice? And if you tag with `Events.tag` at the top, how come you get a numeric rate at the last line? Why not use Duration throughout?

lust4life:
> you have fiveMinutesEWMATicker.TickEvery (Ti... but then you have another line with tick? Why do we have an instance of a ticker used twice?

`Pipe.tick` accept a ticker. a ticker define how to folder with each item from `stream/pipe` and handler tick (generate output for next `stream/pipe`) and can be ticked manually by the user who hold the ticker. or automatically on a `timer`. 

generally, most of the time-related pipe are encapsulating them as one like this:
```fsharp
  let inline bufferTime timespan pipe =
    let ticker = BufferTicker ()
    pipe
    |> withTickJob (ticker.TickEvery timespan)
    |> tick ticker
```

because a timer need be managed, so we should append them to pipe, so when shutdown registry, we should stop them, like `ctss` below:
```fsharp
    let rec running ctss =
      Alt.choose [
        ...
        shutdownCh ^=> fun (res, timeout) ->
          rlogger.infoWithAck (eventX "Shutting down")
          ^=>. Seq.Con.iterJob Cancellation.cancel ctss
          >>=. shutdown targets timeout
          >>= fun shutdownInfo -> res *<= shutdownInfo
      ]

    let state =
      { runtimeInfo = ri
        msgProcessing = msgProcessing
        flushCh = flushCh
        shutdownCh = shutdownCh }

    createGlobals conf.runtimeInfo.logger state
    >>=. Seq.Con.mapJob id conf.processing.tickTimerJobs
    >>= fun ctss -> Job.supervise rlogger (Policy.restartDelayed 500u) (running ctss) 
    >>-. state
```

so, first instance usage is for setup a tick timer, second instance usage if for setup a ticker. 

in a situation which user want to control when to tick manually (like when some url route has been matched), they can tick through a ticker, by invoke `ticker.Tick ()`.


> And if you tag with Events.tag at the top, how come you get a numeric rate at the last line? 

`tag` here is a filter, not a setter. here means it filter messages which has tag 'metric request latency'. the message itself should carry the request duration info, here is just a fake example.

lust4life:
> How do would you design the filtering and end-to-end declaration of the processing pipeline?

what does this mean, don't get the point. can you explain that a little bit more?

haf:
I understand; I think the next point for me is to test your code a bit. What state is it in?

lust4life:
- [x] ExpWeightedMovAvg -> events ticker
  - [ ] more unit test
     - [ ] ewma & percentile
     - [x] use pipe builder build an healthchecker
  - [x] support message filter, processing reconfiguration
         can be achieved by create another logary instance, but needs user create logger through static create, if user create log through an logary instance, things does not work for this situation. and we should think about how to stop ticker jobs on pipes, see below.
  - [x] targets rule to event processing
    - [ ] if discussion blow passed, need remove rules on target
  - [ ] event format improve mentioned by #257 
  - [ ] complete todo comments part in codebase
  - [ ] **think about how to GC ticker jobs on events, these may leak mem** ?

the basic `Event/Pipe Processing` part is done, maybe need add more types of pipe (slidingWindowTime, fixWindow num/time...).

just add a test for show how to use Processing to builder an health checker. [ping svc healthchecker](https://github.com/logary/logary/pull/263/commits/826afbc26997c4040801d88fbde7a917e8c35084)

on next, maybe try to add unit test for `ewma & percentile`, the `todo comments` mostly about typesharp for `ValueModule.fs`, it depends on what we discuss about in #257 , at last, maybe need think about how to GC tick jobs when initialize another logary instance, the previous logary instance go out of its lifetime, auto tick jobs need support GC.

lust4life:
> think about how to GC ticker jobs on events, these may leak mem ?

maybe we can support `DisposeAsync` at `LogManager/Registry.T` to solve these problem.

lust4life:
@haf  ping ? 😸

haf:
Hey @lust4life 

You've proven that you're ready to stick to this project and help out, so I'm going to give you commit access; I'm looking forward to you merging this PR (and using your code!).

We're using Logary throughout everything at my company, so I'm absolutely not abandoning this project, but I may have to keep a slightly lower profile. Let's keep the discussions going none the less.

What do you say? Are you up for finishing up this PR and merging it to master as v5?

Cheers

lust4life:
@haf  Last few days i was on holiday in my hometown. Next, I will follow up to finish it. If there are any problems, i will ping you 😸

lust4life:
```fsharp
module Registry =
  /// The holder for the channels of communicating with the registry.
  type T =
    private {
      runtimeInfo : RuntimeInfo
      msgProcessing : Message -> Middleware option -> Alt<unit>

      /// Flush all pending messages from the registry to await shutdown and
      /// ack on the `ackCh` when done. If the client nacks the request, the
      /// `nack` promise is filled with a unit value. Optional duration of how
      /// long the flush 'waits' for targets before returning a FlushInfo.
      flushCh : Ch<Ch<FlushInfo> * Promise<unit> * Duration option>

      /// Shutdown the registry in full. This operation cannot be cancelled and
      /// so the caller is promised a ShutdownInfo.
      shutdownCh : Ch<IVar<ShutdownInfo> * Duration option>
    }

    member private x.Dispose (disposing : bool) =
      (x :> IAsyncDisposable).DisposeAsync () |> Hopac.start

    override x.Finalize () = x.Dispose false

    interface IDisposable with
      override x.Dispose () = 
        x.Dispose true
        GC.SuppressFinalize(x)
        

    interface IAsyncDisposable with
      override x.DisposeAsync () =
        x.shutdownCh *<+=>- fun reply -> reply, None
        |> Job.startIgnore
```

>  maybe need think about how to GC tick jobs when initialize another logary instance ,and the previous logary instance go out of its lifetime, these auto tick jobs need stop.

@haf  should we support something like this ?  or let the users shutdown it manually, don't rely on GC. Because `DisposeAsync` in `IDisposable.Dispose` is not very robust.

lust4life:
Work continues at #263

PR #265, opened 2017-09-08
Error in "Hello World (F#)" example

AkosLukacs:
Copy & paste and got this error message:

```F#
Type mismatch. Expecting a
    'Random * float -> (Random * float) * Message []'    
but given a
    'Random * float -> (Random * float) * Message list'    
The type 'Message []' does not match the type 'Message list'
```


haf:
Thank you!

PR #272, opened 2017-10-05
Bugfix/make sql tests work

bradder555:


haf:
Thank you @bradder555

haf:
Would you be open to helping us maintain the SQL target going forward? It would be **very** appreciated to have someone who uses it, help us maintain it.

bradder555:
Hi Haf, 
I did make a start on getting the mssql target work, but due to time pressures we set up and are using influxdb.

I am still keen to spend some time on this but it'll be outside of work hours

haf:
How fortunate we also support InfluxDB then ;)

PR #289, opened 2018-04-04
Feature/ingestion + codecs

haf:


src/legacy/Logary.Services.Rutta/Ingestion.TCP.fs
```diff
@@ -0,0 +1,49 @@
...
+    // Aborted socket due to closing process:
+    if isNull frame then () else
+    let bs = Socket.recv receiver
+    // http://api.zeromq.org/4-1:zmq-socket#toc19
+    // A connection was made
+    if bs.Length = 0 then streamRecvLoop receiver next else
+    // printfn "Data received: %A" message
+    use ms = new MemoryStream(bs)
+    use sr = new StreamReader(ms, Encoding.UTF8)
+    while not sr.EndOfStream do
```
haf: Slightly broken, see https://stackoverflow.com/questions/524818/what-is-a-good-method-to-handle-line-based-network-i-o-streams/524954#524954 — which is also broken because it assumes ASCII (which it works for, but for UTF8 we may receive partial lines depending on the package sizes)

PR #290, opened 2018-04-04
First cut at OpsGenie v2

haf:


haf:
https://travis-ci.org/logary/logary/jobs/362487352#L1481

Match null as well

PR #291, opened 2018-04-05
Aspnet Core logging adaptor

lust4life:
default output （aspnet core console）：
image

with logary (literate console):

image


src/Logary.AspNetCore/Logger.fs
```diff
@@ -0,0 +1,68 @@
...
+
+open System
+open Microsoft.Extensions.Logging
+open Logary
+open System.Collections.Generic;
+open Logary.Internals.Aether.Optics
+
+module LoggerAdapter =
+
+  type msLogLevel = Microsoft.Extensions.Logging.LogLevel
```
haf: Use uppercase first letter.

src/Logary.AspNetCore/Logger.fs
```diff
@@ -0,0 +1,68 @@
...
+        if isNull formatter then
+          m |> Message.setField "State" state
+        else
+          let customFormated = formatter.Invoke(state, ex)
+          m |> Message.setField "State" customFormated
+
+
+    interface ILogger with
+      member x.BeginScope<'t> (state: 't) : IDisposable =
+        logManager.beginScope (lazy(box state))
```
haf: Why lazy?

src/Logary.AspNetCore/Logger.fs
```diff
@@ -0,0 +1,68 @@
...
+      | :? IEnumerable<KeyValuePair<string, obj>> as kvSeq ->
+        let messageTpl = kvSeq |> Seq.tryFindBack (fun kv -> kv.Key = "{OriginalFormat}") |> Option.map (fun kv -> kv.Value)
+        let fields = kvSeq |> Seq.map (fun kv -> kv.Value ) |> Array.ofSeq 
+        match messageTpl with
+        | Some tpl -> m |> Message.setEvent tpl |> Message.setFields fields
+        | None -> m |> Message.setField "State" state
+
+      | _ ->
+        if isNull formatter then
+          m |> Message.setField "State" state
```
haf: Lowercase?

src/Logary.AspNetCore/Logger.fs
```diff
@@ -0,0 +1,68 @@
...
+    interface ILogger with
+      member x.BeginScope<'t> (state: 't) : IDisposable =
+        logManager.beginScope (lazy(box state))
+
+      member x.IsEnabled (level: msLogLevel) = 
+        if level = msLogLevel.None then false
+        else mapLogLevel level >= logger.level
+
+      member x.Log<'t> (level: msLogLevel, eventId: EventId, state: 't, ex: exn, formatter: Func<'t,exn,string>) =
+        if not ((x :> ILogger).IsEnabled level) then
```
haf: I think this adds a lot of noise. Why not use the deferred callback method? The `messageFactory` ones? They should not evaluate unless the level is right.

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +79,45 @@ let tests = [
...
+      lg.infoWithAck (eventX "1") |> Hopac.run
+
+      logm.flushPending () |> Hopac.run
+      let outStr = clearStream out
+      Expect.stringContains outStr "scope-1" "shoule have scope-1 as its scope"
+
+      using (logm.beginScope (lazy(box ("scope-2",2)))) (fun _ ->
+        lg.infoWithAck (eventX "2") |> Hopac.run
+
+        logm.flushPending () |> Hopac.run
```
haf: This might deadlock

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +79,45 @@ let tests = [
     do! logm.shutdown ()
   }
 
+  testCaseJob "log with scope" (job {
+    let! (logm, out, error)  = Utils.buildLogManager ()
+    let loggername = PointName.parse "logger.test"
+    let lg = logm.getLogger loggername
+
+    using (logm.beginScope (lazy(box "scope-1"))) (fun _ ->
+      lg.infoWithAck (eventX "1") |> Hopac.run
```
haf: THis might deadlock

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +79,45 @@ let tests = [
...
+
+        logm.flushPending () |> Hopac.run
+        let outStr = clearStream out
+        Expect.stringContains outStr "scope-1" "shoule have scope-1 as its scope as well"
+        Expect.stringContains outStr """["scope-2", 2]""" "shoule have scope-2 as its scope"
+      )
+
+      lg.infoWithAck (eventX "out 2") |> Hopac.run
+
+      logm.flushPending () |> Hopac.run
```
haf: This might deadlock

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +79,45 @@ let tests = [
...
+      let outStr = clearStream out
+      Expect.stringContains outStr "scope-1" "shoule have scope-1 as its scope"
+      Expect.isFalse (outStr.Contains("scope-2")) "shoule not have scope-2 as its scope"
```
haf: Consider using `open Expecto.Flip` for new code you write.

lust4life: get it

src/Logary.AspNetCore/Logger.fs
lust4life: since this api can be used from logary itself (not from aspnet core `ILogger`)，and scope value is useful when log message in this scope are really logged. want to make these value be calculated when some log message are send to target.

src/Logary.Tests/Registry.fs
```diff
@@ -79,7 +79,7 @@ let tests = [
     do! logm.shutdown ()
   }
 
-  testCaseJob "log with scope" (job {
+  ftestCaseJob "log with scope" (job {
```
haf: Remove before merge

haf:
I think it's nice, but have you seen the diff MSFT's web server generates? https://github.com/logary/logary/pull/291/files#diff-2bfb11ab4230093400363f8d70eab088

I'm not sure I want to maintain this paket file in the main repo to be honest. Can we create a separate repository for this monster? :)

haf:
Because of the above reason and because this introduces a 2nd level namespace, I think it would be wiser to have a separate repository for it https://github.com/logary/AspNetCore — @lust4life you have write access there now.

May I suggest we place this in a namespace under `Logary.Targets.AspNetCore`? Rather than `Logary.AspNetCore`?

lust4life:
@haf  how can i add this beta version as dependency if we have not publish logary to nuget ? have any idea ?

haf:
Seems like the build script doesn't push to nuget. If you could add it 
I'll do the push tonight.

lust4life:
@haf  separate is done. namespace changed to : `Logary.Adapters.AspNetCore`  see : https://github.com/logary/AspNetCore

haf:
Nice

PR #295, opened 2018-04-09
span/scope support

lust4life:
need add this to logary first (and release), then aspnet core target in separate repo can reference it.


haf:
Doesn't build

src/tests/Logary.Adapters.Facade.Tests/Program.fs
```diff
@@ -39,6 +40,8 @@ let stubLogManager (message: Message ref) =
       member x.shutdown (fDur,sDur) =
         Alt.always (FlushInfo([],[]),ShutdownInfo([],[]))
       member x.switchLoggerLevel (path, minLevel) = ()
+      member x.wrapScope scope = ()
```
haf: What does wrap scope do?

lust4life: aspnet core has its own external scope provider, when aspnet core logging framework `SetScopeProvider`, we wrap it from our own scope.  

https://github.com/logary/logary/blob/aspnet-logging/src/Logary.AspNetCore/LoggerProvider.fs#L14-L23

```fsharp
  interface ISupportExternalScope with
    member x.SetScopeProvider (scopeProvider: IExternalScopeProvider) =
      m.wrapScope {
        new ILogScope with
          member x.collect () = 
            let mutableStateList = new ResizeArray<obj>()
            do scopeProvider.ForEachScope(new Action<obj,ResizeArray<obj>>(fun data state -> state.Add(data)), mutableStateList)
            mutableStateList |> List.ofSeq
          member x.push lazyData = scopeProvider.Push(lazyData.Value)
          member x.wrap scope = ()
} 
```

with this, other lib use aspnet logging abstraction api logging scoped value can be inject into logary too.

haf:
Ok, but what does it do for non-ASP.Net users?

Isn't this like 'get child logger'?

lust4life:
https://github.com/logary/logary/blob/aspnet-logging/src/Logary/LogScope.fs#L49

for non-asp.net users, generally they don't need invoke this method (`wrapScope`), unless they want to provider their own `LogScope` unlike the default one using `AsyncLocal`.

haf:
Ok, but what is a LogScope? Is it a thread-local-storagre mechanism or ?

lust4life:
i add an abstraction `ILogScope` into logary's `Registry.T.logScope`, the default implementation of it is use [`AsyncLocal`](https://docs.microsoft.com/zh-cn/dotnet/api/system.threading.asynclocal-1?view=netcore-2.1) .  like the doc says: 

> Because the task-based asynchronous programming model tends to abstract the use of threads, AsyncLocal<T> instances can be used to persist data across threads.  

it likes a thread-local-storagre mechanism, and can flow to different threads when using with `async/await` in c#.

lust4life:
seems serilog use different mechanism according to specified symbols   https://github.com/serilog/serilog/blob/dev/src/Serilog/Context/LogContext.cs#L55-L62

haf:
Ok, because "scope" can also mean the scope of:

  - a measurement / timer
  - a logger / sensor / measurement name
  - an operation start/end-pair
  - a zipkin span
  - something IDisposable generally

So perhaps "data slot" or "activity" would be better names.

That said; can't we find an abstraction that fits Hopac as well?

lust4life:
https://github.com/Hopac/Hopac/issues/24
seems we can't, functional style encourage passing dependency explicitly (like what we do before, use apply to generate a new child logger), but this style use [**ambient context**](https://codeblog.jonskeet.uk/2010/11/08/the-importance-of-context-and-a-question-of-explicitness/) , don't know if it works with f# `Async`

`ILogContextHole` ? 😂

haf:
Can't this be an extension method on the interface? Doesn't seem like implementations need to know about any of this.

lust4life:
since every logger belongs to one `Registry` which express as `LogManager`, we support multi registry (logary instace) , so i add it to `LogManager` (the public interface to Logary). 

maybe we can make it into `Global`, all loggers use the same one. but this may lead to some confusing when user use multi registry. (will try some test later)

haf:
What I mean is that; can't this be in a middleware inside a child logger? That way it's still scoped to the logger; and that child logger is already disposable, right? All that's needed is to wrap a new instance of the AsyncLogScope in the middleware closure?

lust4life:
do you mean something like this:
```fsharp
let childLogger = logger.withDataSlot(xxx)
childLoggder.info "xxx"
```

but in scenario like below this will not help:
`serviceA` depend on `serviceB` , each service create their own logger using (`Log.create<SomeType> ()`), when `instanceA.a` invoke `instanceB.b`.  `a` can set some ambient log context data, then `b`'s log message can retrieve these data. there is no logger passing between them.

haf:
It would help if `b` returns an async that is executed in `a`'s async context, right?

lust4life:
@haf  it works 👍

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +79,39 @@ let tests = [
...
+    let dataSlot = DataSlot.useDefault ()
+    let (logm, out, _)  = Utils.buildLogManagerWith (fun r -> r |> Config.middleware (Middleware.fillWithContextSlot dataSlot) ) |> run
+
+    for _ in 1..10 do
+
+      let loggername = PointName.parse "logger.test"
+      let lg1 = logm.getLogger loggername
+
+      let s1 = dataSlot.push (fun _ -> "scope-1")
+      lg1.infoWithBP (eventX "scope1") |> run
```
haf: I'm afraid we can't use 'run' in the unit tests or they might deadlock. Could you make `testCase` a `testCaseJob`?

lust4life: done.

lust4life:
@haf  after checking the span concept (http://opentracing.io/) , if we can support this concept and we can easily hook logary target with someting like ([skywalking](https://github.com/apache/incubator-skywalking), [jaeger](https://github.com/jaegertracing), zipkin... ). what do you think ?

src/Logary.Tests/Program.fs
```diff
@@ -426,6 +426,8 @@ let tests =
               ()
             member x.stop (decider: Duration -> LogLevel): Alt<Promise<unit>> =
               Promise.instaPromise
+            member x.createSpan id fac: Span =
```
haf: Can't we push this into `bisect`?

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +78,69 @@ let tests = [
     do! logm.shutdown ()
   }
 
+  testCaseJob "log with scope" (job {
+    let dataSlot = DataSlot.useDefault ()
+    let! (logm, out, _)  = Utils.buildLogManagerWith (fun r -> r |> Config.middleware (Middleware.fillWithContextSlot dataSlot) )
```
haf: No brackets needed `(`, `)`.

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +78,69 @@ let tests = [
...
+      do s2.Dispose ()
+      lg2.infoWithBP (eventX "2dispose") |> Hopac.startWithActions ignore (latch.Decrement >> start)
+
+      do s1.Dispose ()
+      lg2.infoWithBP (eventX "1dispose") |> Hopac.startWithActions ignore (latch.Decrement >> start)
+
+      do! latch |> Latch.await |> Alt.afterJob logm.flushPending
+
+      let outStr = clearStream out
+      Expect.isRegexMatch outStr (new Regex("""scope1.*?_logary.scope => \["scope-1"\]""", RegexOptions.Singleline)) "shoule have scope-1 value"
```
haf: So why not use the parent-child int128 that openzipkin uses? Strings are so much more expensive?

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +78,69 @@ let tests = [
...
+
+    do! logm.shutdown ()
+  })
+
+  testCaseJob "log with span" (job {
+    let! (logm, out, _)  = Utils.buildLogManager ()
+    let checkSpanId spanId = job {
+      do! logm.flushPending ()
+      clearStream out
+      |> Expecto.Flip.Expect.stringContains "should has spanId" spanId
```
haf: `have`, not `has`. Good that you're using `Expecto.Flip`, but even better if we change the whole file to use it.

src/Logary.Tests/Registry.fs
```diff
@@ -78,12 +78,69 @@ let tests = [
...
+      use span1 = lgb.createSpan rootSpan.id (Message.eventX "some child span 1")
+      let span2 = lgb.createSpan rootSpan.id (Message.eventX "some child span 2")
+      do span2.Dispose ()
+
+      do! checkSpanId "#localhost-svc.1.2"
+
+      for i in 1..2 do
+        use spani = lgb.createSpan span1.id (fun level -> Message.eventFormat(level, "some grand span {taskId}", i) )
+        do! timeOutMillis 500
+        do! lgb.infoWithAck (eventX "some log message" >> setSpanId spani.id)
```
haf: Nice, is this what the child logger/span would perform with a middleware/apply?

src/Logary.Tests/Utils.fs
```diff
@@ -42,15 +48,17 @@ let buildLogManager () = job {
...
-    // |> Config.ilogger (ILogger.Console Error)
     |> Config.target twTargetConf
     |> Config.processing (Events.events |> Events.sink [tname])
     |> Config.disableGlobals
+    |> configFac
     |> Config.build
   return logm, out, error
 }
 
+let buildLogManager () = buildLogManagerWith id
```
haf: Nice

src/Logary/Constants.fs
```diff
@@ -35,6 +35,14 @@ module KnownLiterals =
   [<Literal>]
   let SinkTargetsContextName =  LogaryPrefix +  "sink.targets"
 
+  [<Literal>]
+  let ScopeContextName = LogaryPrefix + "scope"
+
+  [<Literal>]
+  let SpanIdContextName = LogaryPrefix + "spanId"
+  [<Literal>]
+  let SpanInfoContextName = LogaryPrefix + "spanInfo"
```
haf: Have you considered the effect on this for the "getOthers" readers and the stuff I added to MessageModule.Patterns?

src/Logary/DataModel.fs
```diff
@@ -209,6 +209,29 @@ type Message =
     member x.timestampEpochS: int64 =
       x.timestamp / Constants.NanosPerSecond
 
+/// span focus on time scope (scope means it needs explicitly specify its end) which usually used for tracing.
+/// message focus on data context which usually used for logging/metric (since metric can be express by gauge which inclued by message).
+/// one message can belong to a span, it means this log message happened in this span.
+/// span itself contains one message, when logging, it just fill message with some specific info (startTime, endTime, duration, id ...)，
+/// then every target can transform this message into their own backend's implementation (like zipkin,jaeger,opentracing...).
+/// since span can be finished, so they will have a logger and send it's message at that time.
```
haf: Good description. I agree that this is a nice abstraction, but since Finish is async, should disposable also be optionally async (IAsyncDisposable)? I think it should be sync disposable, but then `finish` must be called without a caller (on the Hopac scheduler with a timeout, like `logSimple` does it)

src/Logary/DataModel.fs
```diff
@@ -209,6 +209,29 @@ type Message =
...
+  inherit IDisposable
+
+  /// log its data into registry
+  abstract finish: (Message -> Message) -> Promise<unit>
+  /// the id of this span
+  abstract id: string
+
+type NoopSpan () =
+  interface Span with
+    member x.finish trasn = Promise(())
```
haf: instaPromise.

src/Logary/DataSlot.fs
```diff
@@ -0,0 +1,8 @@
+namespace Logary
+
+open System
+
+type DataSlot =
+  abstract collect: unit -> obj list
+  abstract push: Lazy<obj> -> IDisposable
```
haf: Docs? What intention here?

src/Logary/DataSlot.fs
```diff
@@ -0,0 +1,8 @@
...
+  abstract wrap: DataSlot -> unit
```
haf: Intention?

src/Logary/DataSlotModule.fs
```diff
@@ -0,0 +1,59 @@
...
+[<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
+module DataSlot =
+
+  type internal AsyncLocalDataSlot () =
+    let _currentData = new AsyncLocal<Lazy<obj> list>()
+    do _currentData.Value <- list.Empty
+
+    let getCurrentDataFromAsyncLocal () =
+      let current = _currentData.Value
+      if obj.ReferenceEquals(current, null) then
```
haf: Use `isNull`.

src/Logary/Extensions/LogManager.fs
```diff
@@ -0,0 +1,10 @@
+namespace Logary
+
+open System.Runtime.CompilerServices
+
+[<AutoOpen; Extension>]
+module LogManagerEx =
```
haf: Nice.

src/Logary/Internals/InternalLogger.fs
```diff
@@ -1,14 +1,13 @@
...
-open System
 open Hopac
 open Hopac.Infixes
 open Logary
 open Logary.Internals
 open Hopac.Extensions
 
 [<CompilationRepresentation(CompilationRepresentationFlags.ModuleSuffix)>]
-module InternalLogger =
+module internal InternalLogger =
```
haf: Yeah, good. We could also consider internalising more in `Logary.Formatting` as it's a bit hairy to navigate for new users.

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+open System.Collections.Concurrent
+open Logary.Internals
+open Hopac
+open Hopac.Infixes
+
+/// describe the time scope info about a span
+type SpanInfo =
+  {
+    id : string
+    beginAt: int64 // UnixTimeTicks
```
haf: UnixTime = TicksSinceEpoch?

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+open NodaTime
+open System.Collections.Concurrent
+open Logary.Internals
+open Hopac
+open Hopac.Infixes
+
+/// describe the time scope info about a span
+type SpanInfo =
+  {
+    id : string
```
haf: Consider making this int128 like in opentracing.

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+    beginAt: int64 // UnixTimeTicks
+    endAt: int64 // UnixTimeTicks
+    duration: int64
```
haf: Unit?

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+
+          x.logger.logWithAck Info composedMsgFac >>=* id
+        else Promise (())
+
+    interface IDisposable with
+      member x.Dispose () =
+         (x :> Span).finish id |> Hopac.start
+
+
+  let createSpanT (clock: IClock) (idGen: SpanIdGenerator) (parentSpanId: string) (logger: Logger) (msgFac: LogLevel -> Message) : T =
```
haf: Rename to `create`.

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+              endAt = endAt.ToUnixTimeTicks()
+              duration = dur.BclCompatibleTicks
+            }
+
+            (x.messageFac >> transform) level
+            |> Message.setContext KnownLiterals.SpanInfoContextName spanInfo
+            |> Message.setSpanId x.id
+
+          x.logger.logWithAck Info composedMsgFac >>=* id
+        else Promise (())
```
haf: `instaPromise`

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+  /// span itself contains one message, when logging, it just fill message with some specific info (startTime, endTime, duration, id ...)，
+  /// then every target can transform this message into their own backend's implementation (like zipkin,jaeger,opentracing...).
+  /// since span can be finished, so they will have a logger and send it's message at that time.
+  type T =
+    {
+      messageFac: LogLevel -> Message
+      id: string
+      beginAt: Instant
+      logger: Logger
+      hasFired: bool
```
haf: What does this mean? That it's logged?

src/Logary/Internals/SpanModule.fs
```diff
@@ -0,0 +1,99 @@
...
+open System.Threading
+open System
+open NodaTime
+open System.Collections.Concurrent
+open Logary.Internals
+open Hopac
+open Hopac.Infixes
+
+/// describe the time scope info about a span
+type SpanInfo =
```
haf: Consider making it a struct type.

src/Logary/Registry.fs
```diff
@@ -25,62 +26,12 @@ type LogaryConf =
...
-  /// this will only affect the loggers (its name, not its instance) which have been created beafore
-  abstract switchLoggerLevel: string * LogLevel -> unit
-
 /// This is the main state container in Logary.
 module Registry =
   /// The holder for the channels of communicating with the registry.
   type T =
     private {
-      runtimeInfo: RuntimeInfo
+      runtime: RuntimeInfo
```
haf: This is a breaking change; which is fine right now, but let's keep these to a minimum going forward.

haf:
Hi,

Right, but these are two different things:

  1. Identifying the path/location/hierarchy
  2. Identifying the span

Don't we want to identify the span?

src/Logary/Constants.fs
lust4life: do you think we need to add these two literals ? or we just treat them as other normal user defined context value ?

src/Logary/DataModel.fs
lust4life: done. finish now is sync, use `logWith`(same way as logSimple).

src/Logary/DataModel.fs
lust4life: refactor this to sync

src/Logary/DataSlot.fs
lust4life: removed, it's the old implementation.

src/Logary/DataSlot.fs
lust4life: removed, it's the old implementation

src/Logary/DataSlotModule.fs
lust4life: removed, it's the old implementation

src/Logary/Internals/SpanModule.fs
lust4life: yes, it mean this span has been logger (finish its life time).

src/Logary/Internals/SpanModule.fs
lust4life: done.

lust4life:
I didn't check the store layer about zipkin/jaeger..., don't know if they are used for identify the span (through zipkin issue I post above ,seems this is a primary key). will do some investigate later.

Just think of the span as an ordinary message with some more time concern info, since message has no identify id, so i think we should only need it's path ?

lust4life:
you are right, make it simpler. (path/location/hierarchy) is not the span's concern.

src/Logary/DataModel.fs
```diff
@@ -220,18 +220,17 @@ type Span =
...
-    member x.finish trasn = ()
-    member x.id = String.Empty
-
-  interface IDisposable with
-    member x.Dispose () = ()
-
-  static member instance = new NoopSpan () :> Span
+  /// the data of this span
+  abstract info: SpanInfo
+and SpanInfo =
```
haf: This type does not need to be recursive (`and` -> `type`).

src/Logary/Internals/SpanModule.fs
```diff
@@ -3,83 +3,74 @@ namespace Logary
...
+  [<Struct>]
+  type internal SpanLog =
     {
+      traceId: string
+      spanId: string
       parentSpanId: string
-      messageFac: LogLevel -> Message
-      clock: IClock
-      localRootPrefix: string
+      beginAt: int64 // number of ticks since the Unix epoch. Negative values represent instants before the Unix epoch. (from NodaTime)
```
haf: We use NanosSinceEpoch elsewhere; let's not conflate them, please. Ticks <> nanos, but a .Net-specific thing that is 10 000 * something I can never remember (millis?).

src/Logary/Internals/SpanModule.fs
```diff
@@ -3,83 +3,74 @@ namespace Logary
...
 
-  type private SpanBookmark =
+  type private T =
     {
-      spanId: string
-      childCount: int64 ref
+      traceId: Guid
+      spanId: Guid
+      parentSpanInfo: SpanInfo option
+      messageFac: LogLevel -> Message
```
haf: Remember that functions aren't serialisable by default, if you ever need `T` to be.

src/Logary/Internals/SpanModule.fs
```diff
@@ -3,83 +3,74 @@ namespace Logary
...
+            spanId = t.spanId
+          }
         else
-          // root span from this service, prepend local info
-          sprintf "%s#%s.%x" parentSpanId t.localRootPrefix (Interlocked.Increment(localRootCount))
-
-    member x.withParentId pid =
-      do t <- {t with parentSpanId = pid}
+          // has ambient span
+          {
```
haf: Keep with the formatting of no newline after `{`, please.

src/Logary/Internals/SpanModule.fs
```diff
@@ -90,38 +81,32 @@ module SpanBuilder =
...
-          member x.id = spanId
+          member x.info = spanInfo
 
           member x.finish (transform: Message -> Message) =
             let composedMsgFac level =
               let endAt = t.clock.GetCurrentInstant()
               let dur = endAt - beginAt
-              let spanInfo = {
-                id = spanId
+              let spanLog = {
```
haf: What's `spanLog`?

src/Logary/Internals/SpanModule.fs
```diff
@@ -90,38 +81,32 @@ module SpanBuilder =
...
+      let spanInfo = generateSpanInfo ()
       let beginAt = t.clock.GetCurrentInstant()
 
       let previous = activeSpan.Value
-      let bookMark = {
-        spanId = spanId
-        childCount = ref 0L
-      }
-      activeSpan.Value <- bookMark
+      activeSpan.Value <- spanInfo
```
haf: I don't like global statics at all... Globals go in the `Global` module, nowhere else, if possible.

src/Logary/MiddlewareModule.fs
```diff
@@ -65,6 +65,7 @@ module Middleware =
 
   let useAmbientSpanId () : Middleware =
```
haf: What difference is it to a consumer if it's a `span` or a `spanId`? Doesn't "use global span id" imply "use global span"? What's a global span? If I run a web app with 100 concurrent requests on Suave, what's the global span?

haf:
- https://github.com/openzipkin/zipkin/issues/1243
- https://github.com/openzipkin/zipkin/issues/1244

Also see my e-mail to you

haf:
I think will will be easier to do in the async case if we fix #310 and thus the buffering/CEP engine as we can make it a processing step and only log things are think are relevant.

src/Logary/SpanModule.fs
```diff
@@ -0,0 +1,137 @@
...
+
+          member x.finish (transform: Message -> Message) =
+            let composedMsgFac level =
+              let endAt = t.clock.GetCurrentInstant()
+              let dur = endAt - beginAt
+              let spanLog = {
+                traceId = SpanInfo.formatId spanInfo.traceId
+                spanId =  SpanInfo.formatId spanInfo.spanId
+                parentSpanId = spanInfo.parentSpanId |> Option.map SpanInfo.formatId |> Option.defaultValue null
+                beginAt = beginAt.ToUnixTimeTicks()
```
haf: Needs fixing

src/Logary/SpanModule.fs
```diff
@@ -0,0 +1,137 @@
...
+      let previous = activeSpan.Value
+      activeSpan.Value <- spanInfo
+
+      { new Span with
+          member x.info = spanInfo
+
+          member x.finish (transform: Message -> Message) =
+            let composedMsgFac level =
+              let endAt = t.clock.GetCurrentInstant()
+              let dur = endAt - beginAt
```
haf: Better to start a Stopwatch here.

haf:
@lust4life I've rebased and merged this with master, now... I'm open to merging this feature and testing it a bit in my own code; what do you say?

PR #297, opened 2018-04-12
port aliyun target to net core

lust4life:


lust4life:
@haf  cloud you help release this nuget pkg (https://www.nuget.org/packages/Logary.Targets.AliYun/) ?

src/targets/Logary.Targets.Stackdriver/App.config
```diff
@@ -12,6 +12,21 @@
     <gcServer enabled="true" />
 
   <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+  <dependentAssembly>
```
haf: Please remove this change to the StackDriver taget.

haf:
Hi, yes I will but please revert the change to the stackdriver appconfig.

src/targets/Logary.Targets.Stackdriver/App.config
lust4life: done.

PR #324, opened 2018-05-17
Add binding redirect for Newtonsoft.Json

akselsson:
The stackdriver target failed for us due to an assembly mismatch for newtonsoft json. This commit fixes that


haf:
Thank you for your PR and congratulations to the first PR you're making to this repo 👍

Issue #264, opened 2017-08-30
DB.Migrations doesn't always use the correct schema


btrepp:
For SQLServer in
Logary.Targets.DB.Migrations/Migrations.fs

```
 base.Create
      .Table(Defaults.GaugesTable).InSchema(Defaults.Schema)

      .WithColumn("Value").AsDouble().NotNullable()
        .WithColumnDescription("The value of the measure.")

      |> SharedBetweenTables.messageStructure

    |> ignore

    // if you're just shipping
    base.Create.Index("IX_Gauges_EpochNanos").OnTable(Defaults.GaugesTable)
      .OnColumn("EpochNanos").Descending().WithOptions().NonClustered()
    |> ignore
```

The gauges table gets created with the "Logary" schema, but then an index is added without specifying the schema (this defaults to dbo)

Which gives, and leaves a logary schema with a single table VersionInfo

```
[+] Beginning Transaction
[+] Committing Transaction
[+] Beginning Transaction
[10:34:26 INF] /* CreateSchema Logary */
[10:34:26 INF]

[10:34:26 INF]

[10:34:26 INF]

[+] Committing Transaction
[10:34:26 INF] /* VersionSchemaMigration migrating ========================================== */
[10:34:26 INF]

[10:34:26 INF] /* VersionSchemaMigration migrated */
[10:34:26 INF]

[10:34:26 INF]

[10:34:26 INF] /* VersionMigration migrating ================================================ */
[10:34:26 INF]

[10:34:26 INF]

[10:34:26 INF] /* CreateTable VersionInfo */
[10:34:26 INF]

[10:34:26 INF]

[10:34:26 INF] /* VersionMigration migrated */
[10:34:26 INF]

[10:34:26 INF]

[+] Beginning Transaction
[10:34:27 INF] /* VersionUniqueMigration migrating ========================================== */
[10:34:27 INF]

[10:34:27 INF]

[10:34:27 INF] /* CreateIndex VersionInfo (Version) */
[10:34:27 INF]

[10:34:27 INF]

[10:34:27 INF] /* AlterTable VersionInfo */
[10:34:27 INF]

[10:34:27 INF]

[[+] Committing Transaction
10:34:27 INF]

[10:34:27 INF] /* CreateColumn VersionInfo AppliedOn DateTime */
[10:34:27 INF]

[10:34:27 INF]

[10:34:27 INF] /* VersionUniqueMigration migrated */
[10:34:27 INF]

[+] Beginning Transaction
[...]

PR #325, opened 2018-05-18
Defines schema for sql migrations

btrepp:
This ensures that all tables are working in the default/supplied schema.
Before some where being migrated into a default schema, and some into
logary. Depending on the database this may work, but it definitely fails
on sql server.

Also adds a up/down test that works on a localhost SQL Server instance
to ensure that it still works.
This is commented to not break existing infrastructure

Should fix with #264


haf:
I see, yes probably. You're right.

PR #327, opened 2018-05-20
Present Log4J throwables as string

akselsson:
Previously, Log4J throwable were parsed into an array. This made it really hard to read stacktraces in Stackdriver - to read a stacktrace, you had to click twice for each line of the stacktrace.

This commit removes the stacktrace parsing and instead presents the throwable as a raw string

Before:
before

After:
after


haf:
Ok, it makes sense; perhaps going forward we can provide a boolean to the codec to opt in to parsing stacktraces?

Issue #351, opened 2018-06-13
Planned license-related changes


haf:
- [x] Move Facade v4 to Apache 2.0 for use with libraries
 - [x] Specify license for Rutta;  GNU (never linked/used inside app)?
 - [x] logary-js; MIT
 - [x] Logary proper; Change wording to make it free for F# web servers and app frameworks. In the end, it's free if the software runs without complaining. If license check evolves and it was free before evolution, e-mail me and you get a license for free.
    [ ] EULA like Commons Clause but for _ usage in any commercial activity_
 - [x] Improve wording away from "general design" in license; by moving to BSD3 license to license commercially under; triggered by above.
 - [x] Add a clear CLA for contributions/pull requests
 - [x] Add license FAQ to LICENSE.md.
 - [x] Make support options explicit as extra add-on to commercial license.
 - [ ] ~Create a web site; must be easy to maintain~
 - [ ] ~Create a dotnet core template showcasing a full setup with Suave, Expecto, Influx, Grafana~
 - [x] Enable `requireLicenseAcceptance` in nuget package
 - [x] Go through all code that is due for relicensing;
    * Verify which bits need attribution explicitly under Apache 2.0.
 - [x] Make it [even clearer](https://github.com/haf/expecto/issues/256#issuecomment-399021028) how the Facade can connect to things

Timeline: for v5 GA release.

 - `src/Logary`: dual licensed
 - `README.md` and all docs: [CC SA](https://creativecommons.org/licenses/by-sa/2.5/)
 - `src/Logary.Facade`: Apache 2.0
 - `src/targets`: dual licensed; some are derived work by others; go through git history
 - `src/adapters`: dual licensed
 - `src/services/Rutta`: [GNUv3 Lesser](http://www.gnu.org/copyleft/lesser.html) or purchase commercial
 - `src/tests`: dual licensed
 - `src/logary-js`: dual licensed

Reading:

 - http://www.jus.umu.se/digitalAssets/13/13669_jan_eriksson_leido.pdf
 - https://medium.com/@liftm/commons-clause-isnt-that-bad-df3ea7c93996

ReedCopsey:
>  Logary proper; Change wording to make it free for F# web servers and app frameworks. 

I'm curious how you intend to write this. Even Suave, for example, uses and is built on top of non-F# code/libraries internally (libuv, all of the templating engines, etc).

haf:
[...]

PR #353, opened 2018-06-14
use await instead of result

SimonCropp:
not sure on this one?

It seems like `await` should be used ?


haf:
Thank you! Looking good. I'm going to be distributing licenses for different things and will be adding a CLA that ensures you're good with the license choices in #351.

PR #363, opened 2018-08-12
Format log messages so that Stackdriver error reporting picks up exceptions

akselsson:
Stackdriver error reporting can automatically pick up errors from log
entries if they are formatted properly:

"When logging error data from App Engine, Kubernetes Engine or Compute
Engine, the only requirement is for the log entry to contain the full
error message and stack trace. It should be logged as a multi-line
textPayload or in the message field of jsonPayload"

"C#: Must be the return value of Exception.ToString()"

https://cloud.google.com/error-reporting/docs/formatting-error-messages

This PR aims to change the formatting of errors to conform with those guidelines. 

Note that I considered using tokeniseExceptions from MessageWriter, but that function outputs a stacktrace that is not equal to Exception.ToString() and thus does not conform to Google's recommendations.

### DIFF
--- a/src/targets/Logary.Targets.Stackdriver/Target_Stackdriver.fs
+++ b/src/targets/Logary.Targets.Stackdriver/Target_Stackdriver.fs
@@ -20,7 +20,9 @@ open Logary.Internals.TypeShape.Core
 open Logary.Message
 open Logary.Target
 open Logary.Formatting
+open System.Globalization
 open System
+open System.IO
 open System.Collections.Generic
 open System.Numerics
 open System.Runtime.CompilerServices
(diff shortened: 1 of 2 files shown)


akselsson:
In the future, we may want to add support for supplying serviceContext.service and serviceContext.version to Stackdriver since service is also a required field. I have, however found that it's sufficient to add an application property to jsonPayload instead, which works for us.

haf:
Thank you @akselsson I'll put out a new release with this change in it!

haf:
Fixes https://github.com/logary/logary/issues/349

haf:
Released as v5-beta.22

PR #376, opened 2018-12-06
Feature/influx tags

haf:


haf:
Merged to master in other commit.

Issue #294, opened 2018-04-08
Some FsMessageTemplate tests failing


haf:
I rewrapped the tests properly in as Expecto tests:

<img width="1030" alt="screen shot 2018-04-08 at 16 04 52" src="https://user-images.githubusercontent.com/193115/38468462-ad407c34-3b46-11e8-9e6c-a80b9d72168c.png">

And these are failing:

(examples)
```
[16:03:51 ERR] logary/message templates/alignment/10: cus #{CustomerId,10:000000}, pleasure to see you failed in 00:00:00.
format should work.
          Expected string to equal:
          "cus #    001234, pleasure to see you"
                ↑
          The string differs at index 5.
          "cus #001234    , pleasure to see you"
                ↑
          String does not match at position 5. Expected char: ' ', but got '0'.
  /Users/h/dev/logibit/logary/src/Logary.Tests/FsMessageTemplates.fs(67,1): Logary.Tests.FsMessageTemplates.RenderedAs@67-8.Invoke(AssertException exn)
 <Expecto>
[16:03:51 ERR] logary/message templates/alignment/9: cus #{CustomerId,10}, pleasure to see you failed in 00:00:00.
format should work.
          Expected string to equal:
          "cus #      1234, pleasure to see you"
                ↑
          The string differs at index 5.
          "cus #1234      , pleasure to see you"
                ↑
          String does not match at position 5. Expected char: ' ', but got '1'.
  /Users/h/dev/logary/src/Logary.Tests/FsMessageTemplates.fs(67,1): Logary.Tests.FsMessageTemplates.RenderedAs@67-8.Invoke(AssertException exn)
 <Expecto>
[16:03:51 ERR] logary/message templates/alignment/8: cus #{CustomerId,-10:000000}, pleasure to see you failed in 00:00:00.
format should work.
          Expected string to equal:
          "cus #001234    , pleasure to see you"
                ↑
          The string differs at index 5.
          "cus #    001234, pleasure to see you"
                ↑
          String does not match at position 5. Expected char: '0', but got ' '.
  /Users/h/dev/logary/src/Logary.Tests/FsMessageTemplates.fs(67,1): Logary.Tests.FsMessageTemplates.RenderedAs@67-8.Invoke(AssertException exn)
 <Expecto>
```

[...]

Issue #322, opened 2018-05-08
Optimise for high load


haf:
Goal: Support shipping 10k messages per second to stackdriver.

Stage 1:

![Before](https://user-images.githubusercontent.com/193115/39761918-a01fe79c-52d9-11e8-95f7-a7f1b89b89c4.PNG)

 - [x] Remove closure allocation in logWithTimeout (by synchronising immediately on buffer)
    - Should remove one Alt.choose, one timeout item Alt

![After](TODO)

Stage 2:

 - [ ] Fix #310 and see what optimisations are available

haf:
```
BenchmarkDotNet=v0.10.14, OS=Windows 10.0.14393.2068 (1607/AnniversaryUpdate/Redstone1)
Intel Core i7-7920HQ CPU 3.10GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
Frequency=3023437 Hz, Resolution=330.7494 ns, Timer=TSC
.NET Core SDK=2.1.100
  [Host]     : .NET Core 2.0.5 (CoreCLR 4.6.26020.03, CoreFX 4.6.26018.01), 64bit RyuJIT DEBUG
  Job-TVKGOY : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.2563.0
  Job-LLLTYO : .NET Core 2.0.5 (CoreCLR 4.6.26020.03, CoreFX 4.6.26018.01), 64bit RyuJIT

Concurrent=True  Server=True  InvocationCount=128  
LaunchCount=1  WarmupCount=1024  

                Method | Runtime |     Mean |     Error |     StdDev |   Median | Rank |
---------------------- |-------- |---------:|----------:|-----------:|---------:|-----:|
 LogHelloWorld_WithAck |     Clr | 250.6 ns |  5.053 ns |   7.084 ns | 247.7 ns |    2 |
  LogHelloWorld_WithBP |     Clr | 156.1 ns |  3.137 ns |   4.883 ns | 155.8 ns |    1 |
  LogHelloWorld_Simple |     Clr | 293.7 ns |  6.522 ns |  10.715 ns | 289.9 ns |    3 |
 LogHelloWorld_WithAck |    Core | 519.0 ns | 90.334 ns | 266.351 ns | 433.2 ns |    4 |
  LogHelloWorld_WithBP |    Core | 319.0 ns | 46.788 ns | 137.954 ns | 280.9 ns |    3 |
[...]

PR #378, opened 2018-12-14
fix pipe compose and rethink the log error model

lust4life:
- fix pipe compose

  `Events.build` here means dispatch each event/message to all pipes, not chains them.

- use message as the error model for logging error

  since this can express more situations and can take with more info. if some error occurs, we can print the message to see if it is registry closed, or timeout, or partially success (and show which targets/message failed).

- support timeout the right way

  each put message to buffer can have it's own waiting time, if timeout, giving up putting messages return a timeout error message to the client. But this doesn't give putting message to all targets a  transaction semantic meaning.

TODO:

- [x] add a logging error handler for registry which can be configured by users to check their targets's pressure / pipe processing's health.
- [x] fix `LoggerModule`
  exposing a friendly api to help client handle logging errors
- [x] fix/add test parts for this refactor changes
- [x] fix other projects error affected by this breaking changes

fixed: #294  #303


lust4life:
@haf  hi,
Doing some changes about the logging error model and try offer something like waiting time for putting message to buffers. Can you help to see if this makes sense.

And i am a litter bit confusing about this `logAck` method's implementation,  it will start logging immediately, did we do this on purpose?

```fsharp
  /// Special case: e.g. Fatal messages.
  let logAck (logger: Logger) level messageFactory: Promise<unit> =
    let ack = IVar ()
    let inner =
      logger.logWithAck (true, level) messageFactory ^=> function
        | Ok promise ->
          Job.start (promise ^=> IVar.fill ack)
        | Result.Error Rejected ->
          IVar.fill ack ()
        | Result.Error (BufferFull target) ->
          //let e = exn (sprintf "logWithAck (true, _) should have waited for the RingBuffer(s) to accept the Message. Target(%s)" target)
          //IVar.fillFailure ack e
          IVar.fill ack ()
    start inner
    ack :> Promise<_>
```

or how about this ?

```f#
  /// log start when user read the promise returned by this method
  let logAckCold (logger: Logger) level messageFactory: Promise<unit> =
    logger.logWithAck (true, level) messageFactory
    >>=* function
    | Ok promise -> promise
    | Result.Error e -> Promise.unit
```

src/Logary/Configuration/Events.fs
```diff
@@ -31,23 +31,33 @@ module Events =
...
-        |> Job.map (function
-          | Ok _ ->
-            LogResult.success
-          | Result.Error [] ->
-            failwithf "No results for %A" (List.ofSeq msgs)
-          | Result.Error (e :: _) ->
-            Alt.always (Result.Error e))
-        )
+        |> Seq.Con.mapJob (fun msg -> next msg |> PipeResult.orDefault LogResult.success)
+        >>= IVar.fill putAllPromises
```
haf: Why not just bind `reduce` as the callback of `Seq.Con.mapJob`'s return value? Why start in parallel and join the result with a promise?

src/Logary/Configuration/Ticker.fs
```diff
@@ -28,11 +29,11 @@ type Ticker<'state,'t,'r> (initialState:'state) =
   member this.Ticked = tickCh :> Alt<_>
   member this.Tick () = tickCh *<- ()
 
-  member this.TickEvery timespan =
+  member this.TickEvery (duration: Duration) =
     let cancellation = Cancellation.create ()
     let rec loop () =
       Alt.choose [
-        timeOut timespan ^=> fun _ ->
+        timeOut (duration.ToTimeSpan()) ^=> fun _ ->
```
haf: You need to guard the boundaries here, since e.g. Duration.Zero and Duration.MaxValue (??) may cause runtime exceptions.

src/Logary/DataModel.fs
```diff
@@ -275,25 +275,8 @@ and SpanInfo =
   static member formatId (id: Guid) =
     id.ToString("n")
 
-/// Why was the message/metric/event not logged?
-[<Struct>]
-type LogError =
```
haf: Thing is, if we change this, there's a whole heap of work to do in the Facade as well. I'm OK with having this **very** simple, like now, because we have only a few cases that happen:

 - target fails terminally due to target's state (e.g. out of disk)
 - target fails grey due to e.g. due to too slow processing 
 - connectivity broken / DNS lookup failure, etc; all inside the logary target, and those may be cached, so a service restart is needed
 - rate of logging > rate of sending, dropping in part
 - misconfiguration

We don't really want to have timeouts in the pipeline if we can avoid it, because every time we add another timeout, it's a work item that Hopac's scheduler has to track. If the Hopac scheduler's queue can't keep up, it means the buffers will fill up, which means we'll run out of memory and crash due to that (since we still input messages into the buffer pipeline); that was the reason of the refactor this summer, away from 'log simple' having a timeout alternative, towards this.

src/Logary/DataModel.fs
```diff
@@ -302,11 +285,13 @@ type Logger =
...
-  /// - `waitForBuffers`: This causes the logger/caller to block on the RingBuffer being
-  ///   available to take the message. Giving a true here may cause the
-  abstract logWithAck: waitForBuffers:bool * level:LogLevel -> messageFactory:(LogLevel -> Message) -> LogResult
-
-  //abstract log: Message -> Alt<Result<Promise<unit>, unit>>
+  /// Returns an Alt that commits on ALL N Targets' buffers accepting the message.
+  /// Even if the Alt was not committed to, one or more targets (fewer than N) may have accepted the message.
+  /// And this can not be canceled.
+  ///
+  /// - `putBufferTimeOut`: time out duration waiting for each target buffer to be available to take the message.
```
haf: I don't think we should have a buffer timeout at all; if it can't take the message, we must reject the message (optionally logging that internally!)

src/Logary/Internals/NullLogger.fs
```diff
@@ -8,7 +8,7 @@ open Logary
 /// itself.
 type private NullLogger() =
   interface Logger with
-    member x.logWithAck (_, _) _ = LogResult.success
```
haf: Why move this? Now it's a three-object allocation every call, as opposed to re-using from the deleted module above, which doesn't allocate anything.

src/Logary/LogResultModule.fs
```diff
@@ -0,0 +1,38 @@
...
+  let error message : LogResult = Alt.always (Result.Error message)
+
+  let notAcceptByTarget (targetName: string) = error (Message.eventFormat("not accept by {target}", targetName))
+
+module internal LogError =
+  let targetBufferFull (targetName: string): ProcessResult = Result.Error (Message.eventFormat("{target} 's log buffer is full", targetName))
+
+  let timeOutToPutBuffer (targetName: string) (waitSeconds: double): ProcessResult = Result.Error (Message.eventFormat("time out: waiting {duration} seconds to putting message on to {target} 's log buffer", waitSeconds, targetName))
+
+  let clientAbortLogging: ProcessResult = Result.Error (Message.eventFormat("client abort logging"))
```
haf: This result would never go anywhere; is this what you meant with transactional?

src/Logary/LogResultModule.fs
```diff
@@ -0,0 +1,38 @@
...
+module internal LogResult =
+  open Hopac
+
+  let success: LogResult = Alt.always (Result.Ok Promise.unit)
+  let error message : LogResult = Alt.always (Result.Error message)
+
+  let notAcceptByTarget (targetName: string) = error (Message.eventFormat("not accept by {target}", targetName))
+
+module internal LogError =
+  let targetBufferFull (targetName: string): ProcessResult = Result.Error (Message.eventFormat("{target} 's log buffer is full", targetName))
```
haf: We don't want to allocate if we're full if we can avoid it. (generally; this is not a comment on having Message values in the Result)

src/Logary/LogResultModule.fs
```diff
@@ -0,0 +1,38 @@
...
+
+  let timeOutToPutBuffer (targetName: string) (waitSeconds: double): ProcessResult = Result.Error (Message.eventFormat("time out: waiting {duration} seconds to putting message on to {target} 's log buffer", waitSeconds, targetName))
```
haf: This is very dangerous territory; we've designed the buffers to either return true or false; if they time out, it means the scheduler is overcrowded, which means we'll probably crash. We don't want to allocate then, nor will anything be available to actually log/write this message, since nothing gets scheduled anew.

src/Logary/LoggerModule.fs
```diff
@@ -2,61 +2,39 @@
...
-    logger.logWithAck (false, logLevel) messageFactory ^-> function
-      | Ok _ ->
-        true
-      | Result.Error Rejected ->
-        true
-      | Result.Error (BufferFull _) ->
-        false
-
-  let private printDotOnOverflow success =
-    if not success then System.Console.Error.Write '.' else ()
```
haf: This is/was also a good signal on when another Hopac job crashed the scheduling of tasks; by crashing one of the worker threads. If we don't print something here, we'll never be able to see this error case from the logs, since Logary is now indisposed.

src/Logary/Registry.fs
```diff
@@ -130,15 +127,15 @@ module Registry =
       { new Logger with
           member x.name = name
 
-          member x.logWithAck (waitForBuffers, level) messageFactory =
+          member x.logWithAck (putBufferTimeOut, level) messageFactory =
             if level >= x.level then
               // When the registry is shut down, reject the log message.
-              let rejection = t.isClosed ^->. Result.Error Rejected
+              let rejection = t.isClosed ^->. LogError.registryClosed
```
haf: Nice

src/Logary/Registry.fs
```diff
@@ -267,7 +265,18 @@ module Registry =
...
+          let putBufferTimeOut = msg |> Message.tryGetContext KnownLiterals.WaitForBuffers |> Option.defaultValue Duration.Zero
+          msg 
+          |> Target.logAllReduce putBufferTimeOut targets
+          |> Alt.afterFun (fun result ->
+            start (Job.thunk (fun _ ->
+              try
+                conf.logResultHandler result
+              with
+              | e ->
+                 eprintfn "%A" e ))
```
haf: I think `%O` would be better here as all exceptions implement `ToString`.

src/Logary/TargetModule.fs
```diff
@@ -26,95 +28,45 @@ module Target =
...
   ///
-  /// Returns an Alt that commits on the Targets' buffer accepting the message OR the RingBuffer telling the caller that
-  /// it was full, and returning an Error(BufferFull [targetName]) LogResult.
-  let tryLog (x: T) (msg: Message): LogResult =
+  /// - `putBufferTimeOut` means it will try to wait this duration for putting the message onto buffer,otherwise return timeout error, and cancel putting message.
+  ///   if `putBufferTimeOut <=  Duration.Zero` , it will try to detect whether the buffer is full and return immediately
+  let log (putBufferTimeOut: Duration) (x: T) (msg: Message): LogResult =
     let msg = x.transform msg
-    if not (x.accepts msg) then LogResult.rejected
+    if not (x.accepts msg) then LogResult.notAcceptByTarget x.name
```
haf: `not accept` -> `not accepted`

src/Logary/TargetModule.fs
```diff
@@ -26,95 +28,45 @@ module Target =
...
+        RingBuffer.tryPut x.api.requests targetMsg ^-> function
         | true ->
           Result.Ok (upcast ack)
         | false ->
-          Result.Error (BufferFull x.name)
+          LogError.targetBufferFull x.name
+      else
+        RingBuffer.put x.api.requests targetMsg ^->. Result.Ok (upcast ack)
+        <|>
+        timeOut (putBufferTimeOut.ToTimeSpan()) ^-> fun _ -> LogError.timeOutToPutBuffer x.name putBufferTimeOut.TotalSeconds
```
haf: This is what caused our cascading failures of logging high-throughput at tradera.com. We can't go back to it.

haf:
`logAck`:

> And i am a litter bit confusing about this logAck method's implementation, it will start logging immediately, did we do this on purpose?

I'm not sure without reading its call-sites. I'll have to go through them. But generally, you're right; it's a hot function and should probably not be unless a call site specifically requires it to be.

src/Logary/Configuration/Ticker.fs
lust4life: will do it later

src/Logary/DataModel.fs
lust4life: there is a litter difference here, in old one: message will not be dropped when the timeout alternative committed, since we have more than one targets. but right now, if timeout, then message will not be append to the buffer, and we will get a timeout error message to return to the client. if we config a global logging error handler on the registry, we can notify this situation, then adjust the target buffer length ,or speed up the processing. 

since we have a timeout protocol, user can do backpressure with some balance, no need to wait all the time (since logging is not such important as the business code), but we still want to give some time to buffer if they can't catch up in a short time.

src/Logary/DataModel.fs
lust4life: if the timeout is `zero`, it act like the old `trylog`, it will return a timeout error immediately, and the registy's `logResultHandler` can notify this situation.

src/Logary/Internals/NullLogger.fs
lust4life: you are right, this can be avoid, this change is because `LogResultModule` is in a lower file order, `NullLogger` can not reference.

src/Logary/LogResultModule.fs
lust4life: the logging error message return by the logging is not designed by resend to the buffer, it gives a chance to the client to notify the error details just like the old DU `LogError`, and users can choose a safe way to get the notification by the error handler.

src/Logary/LoggerModule.fs
lust4life:
```f#
          msg 
          |> Target.logAllReduce putBufferTimeOut targets
          |> Alt.afterFun (fun result ->
            start (Job.thunk (fun _ ->
              try
                conf.logResultHandler result
              with
              | e ->
                 eprintfn "%A" e ))
```

this has been implement by registry logging error handler in here https://github.com/logary/logary/blob/fix-pipe-compose/src/Logary/Configuration/Config.fs#L49

```f#
 logResultHandler = function | Result.Error error -> System.Console.Error.Write (MessageWriter.singleLineNoContext.format error) | _ -> ()
```

src/Logary/TargetModule.fs
lust4life: do we have some perf-test to check this situation?  like this one ? #322

src/Logary/LogResultModule.fs
haf: Can we at least make it non-allocating, besides the insertion of the variables into an object?

src/Logary/TargetModule.fs
haf: Yes, let's benchmark it.

haf:
This PR, given that it changes public API, needs to update the facade, and the facade adapter to match.

I think you've done a good job, but I would prefer to have the bool in place, and not change the public API; we can make the wait duration configurable with an environment variable instead.

src/Logary/DataModel.fs
lust4life: https://github.com/logary/logary/blob/6b049fa05534bb55b172c237f09e2f16eec5c851/src/Logary/TargetModule.fs#L42

it has been considered. see here

lust4life:
yes, you are right. i have notice that it will affect facade and its adapter. maybe we can add an extension method to add this duration in message context, not change the origin public api, if waitforbuffer is true, we can check whether we have this duration timeout.

lust4life:
@haf  fix adapter and facade. In facade I change the `LogError` to `string`, since facade here is in beta, do I need add a v5 version and do adapter for v4 (`LogError`) ? or can I simply change `LogError` to `string` to hotfix here.  or we need keep `LogError` DU in facade (means no need change facade), change adapter to handle this situation ?

In logary at this refactor, use `Message` instead of `LogError` here is to be able to express more content in complex scenarios (e.g. `Pipe line processing, a DU can not express all situation and message can take more context info for user debug/repair`)

haf:
This error on appveyor looks a bit like a race condition... Will you check it out, or should I?

```
[09:38:36 ERR] logary/Registry/flush/shutdown timeout and logging after shutdown with no blocking failed in 00:00:00.3260000. 
should have no shutdown ack target. Actual value was ["mockTarget"] but had expected it to be [].
  C:\projects\logary\src\Logary.Tests\Registry.fs(203,1): Logary.Tests.Registry.x2yJ@203-641.Invoke(String x)
 <Expecto>
[09:38:36 INF] EXPECTO! 480 tests run in 00:00:07.6237606 for logary - 469 passed, 10 ignored, 1 failed, 0 errored.  <Expecto>
[09:38:36 INF] EXPECTO?! Summary...
 ```

**Comment by haf** [comment]
> @haf fix adapter and facade. In facade I change the LogError to string, since facade here is in beta, do I need add a v5 version and do adapter for v4 (LogError) ? or can I simply change LogError to string to hotfix here. 

I think we can keep it v4 based on that it's in beta. For my own use-case it's easy enough to switch.

**Comment by haf** [comment]
> In logary at this refactor, use Message instead of LogError here is to be able to express more content in complex scenarios (e.g. Pipe line processing, a DU can not express all situation and message can take more context info for user debug/repair)

Sure

File: examples/Libryy/LoggingV4.fs [hunk 14]
```diff
@@ -360,6 +345,14 @@ type Message =
     |> Seq.map (fun (KeyValue (k, v)) -> k.Substring(Literals.FieldsPrefix.Length), v)
     |> Map.ofSeq
 
+  member x.getContext(): Map<string, obj> =
+    x.context
+    |> Map.toSeq
```
haf: https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/map.filter%5B%27key%2C%27t%5D-function-%5Bfsharp%5D ?

lust4life:
> This error on appveyor looks a bit like a race condition... Will you check it out, or should I?

I really have no idea about what is going on here, I don't know if this has anything to do with hopac scheduling timeout.

since we do `timeOut` here first then push flush request to target, https://github.com/logary/logary/blob/fix-pipe-compose/src/Logary/Registry.fs#L172-L174

lust4life:
@haf  can you help release a beta version ? so that i can fill the gap between https://github.com/logary/AspNetCore and the newest logary

lust4life:
@haf   >  Yes, let's benchmark it.

fix the `Logary.PerfTest` , and try some perf test. 

image

image

image

haf:
Let's make the CI pass, so I can merge it. I'll look at this ASAP; sorry for the delay.

lust4life:
seems that : 
- we should encourage users to use `xxxWithBP`, if targets's processing rate can catch up (or the ringbuffer size is enough long), there's almost no cost . if targets can't catch up, then a user defined timeout will make some back pressure for users.
- do `xxxWichAck` if user want to ensure to get the ack from the backend targets's processing
- I don't think we should wrap an asynchronous api in a sync way, e.g `logSimple/logWith`
  
  the reason here is, if user have `Job/Task` , they can decide the way to start this job : 
  - start from the current thread (blocking) until the real job occurs, but no need waiting for result. e.g. `startAsTask` in tpl or `start the job` in hopac
  - start from the background thread , no blocking either, e.g. `queueAsTask` in tpl , `queue the job` in hopac

and if we expose `logSimple`, i think we should use `queue` instead of `start`, since start here will rearrange the worker thread's workstack, if the worker thread is busy doing something, and users log message in a high rate, this will become slower, `queue` here only append a work into the workstack which means have a constant rate (the perf result above show this), but the drawback is message factory doesn't invoke in the user's call-site. so we should leave this decision to users.

i think, `start` here maybe the real reason to slow the perf, not the `alt.timeout` ?

what do you think ?

haf:
Sounds well reasoned to me.

> I don't think we should wrap an asynchronous api in a sync way;

But often when users don't have access to a synchronisation context they don't want to care, so not exposing this forces something on the user and causes questions they may not need to understand.

We could have configuration availble possible that changes the default when the ring buffer is full, though.

haf:
Releasing this as beta-29 now.

PR #386, opened 2019-02-11
add jaeger target and aspnetcore adapter in sln to make build

lust4life:


haf:
Needs it to compile; is it something that I've broken?

lust4life:
maybe because https://github.com/logary/logary/commit/8b756436f564bee551f6109504d455651da27919#diff-6074983490330a0aac1bd23c958232d7L38 
since fake will use an old [`paket.restore.target`](https://github.com/fsharp/FAKE/blob/release/next/src/app/fake-cli/fake-cli.fsproj#L29) file which has a bug to find the global paket .

lust4life:
seems working

haf:
Nice

PR #387, opened 2019-02-12
[WIP] create a Helm chart for Rutta

haf:
- [x] Initially just run Rutta
- [x] Extend Rutta with tests that all targets' configs can be instantiated
- [ ] Write 'test' for helm chart
  * Test single UDP message + console log
  * Running `rutta test` in order to iterate through all available targets and instantiate their configs (to see if we have assembly binding errors)
- [ ] Run `helm test` as part of CI pipeline
- [ ] Send PR to https://github.com/helm/charts/tree/master/incubator


haf:
Need to fix #388

haf:
You are owner of this too @lust4life — or would you prefer not to field questions about it?

lust4life:
sorry been busy these days, will check it later and give some feedback.

PR #395, opened 2019-02-21
Support prometheus

### LINKED ISSUE #361: Prometheus exporter?
Of health/metrics

 + http://marselester.com/prometheus-on-kubernetes.html

lust4life:
- should fix #361 

todo:
- [x] implement counter base type (`DoubleAdder`) and benchmark/ut
- [x] design gauge/counter/Histograms api
- [x] write exporter
- [x] fix gap between logary's log api


lust4life:
``` ini

BenchmarkDotNet=v0.10.14, OS=macOS 10.14.3 (18D109) [Darwin 18.2.0]
Intel Core i5-7500 CPU 3.40GHz (Kaby Lake), 1 CPU, 4 logical and 4 physical cores
.NET Core SDK=2.2.101
  [Host]     : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT DEBUG
  Job-OWVIPG : .NET Core 2.1.6 (CoreCLR 4.6.27019.06, CoreFX 4.6.27019.05), 64bit RyuJIT

InvocationCount=4  LaunchCount=1  TargetCount=4
UnrollFactor=2  WarmupCount=2

```
|      Method | threadCount |     Mean |      Error |    StdDev | Scaled | ScaledSD | Allocated |
|------------ |------------ |---------:|-----------:|----------:|-------:|---------:|----------:|
| **doubleAdder** |           **1** | **13.06 ns** |  **0.3794 ns** | **0.0588 ns** |   **0.94** |     **0.04** |       **0 B** |
| interLocked |           1 | 13.87 ns |  4.2817 ns | 0.6638 ns |   1.00 |     0.00 |       0 B |
|             |             |          |            |           |        |          |           |
| **doubleAdder** |           **2** | **18.10 ns** | **29.3223 ns** | **4.5458 ns** |   **0.42** |     **0.09** |       **0 B** |
| interLocked |           2 | 43.43 ns |  8.6338 ns | 1.3385 ns |   1.00 |     0.00 |       0 B |
|             |             |          |            |           |        |          |           |
| **doubleAdder** |           **4** | **16.19 ns** | **21.0598 ns** | **3.2649 ns** |   **0.28** |     **0.05** |    **2192 B** |
| interLocked |           4 | 58.39 ns |  6.3111 ns | 0.9784 ns |   1.00 |     0.00 |    2192 B |

lust4life:
``` ini

BenchmarkDotNet=v0.10.14, OS=ubuntu 16.04
Intel Core i9-7980XE CPU 2.60GHz, 1 CPU, 36 logical and 18 physical cores
.NET Core SDK=2.2.104
  [Host]     : .NET Core 2.2.2 (CoreCLR 4.6.27317.07, CoreFX 4.6.27318.02), 64bit RyuJIT DEBUG
  Job-DDEUDR : .NET Core 2.2.2 (CoreCLR 4.6.27317.07, CoreFX 4.6.27318.02), 64bit RyuJIT

InvocationCount=4  LaunchCount=1  TargetCount=4
UnrollFactor=2  WarmupCount=2

```
|      Method | threadCount |       Mean |      Error |     StdDev | Scaled | ScaledSD | Allocated |
|------------ |------------ |-----------:|-----------:|-----------:|-------:|---------:|----------:|
| **doubleAdder** |           **1** |  **12.160 ns** |  **0.0529 ns** |  **0.0082 ns** |   **1.00** |     **0.00** |       **0 B** |
| interLocked |           1 |  12.177 ns |  0.0512 ns |  0.0079 ns |   1.00 |     0.00 |       0 B |
|             |             |            |            |            |        |          |           |
| **doubleAdder** |           **2** |  **24.793 ns** | **89.4642 ns** | **13.8695 ns** |   **0.35** |     **0.17** |       **0 B** |
| interLocked |           2 |  70.519 ns |  5.1232 ns |  0.7942 ns |   1.00 |     0.00 |       0 B |
|             |             |            |            |            |        |          |           |
| **doubleAdder** |           **4** |  **17.777 ns** | **46.1038 ns** |  **7.1474 ns** |   **0.19** |     **0.07** |    **2192 B** |
| interLocked |           4 |  92.325 ns |  6.4076 ns |  0.9934 ns |   1.00 |     0.00 |    2192 B |
|             |             |            |            |            |        |          |           |
| **doubleAdder** |           **8** |   **9.072 ns** |  **6.5144 ns** |  **1.0099 ns** |   **0.08** |     **0.01** |    **3888 B** |
| interLocked |           8 | 107.841 ns | 11.2586 ns |  1.7454 ns |   1.00 |     0.00 |    3888 B |
|             |             |            |            |            |        |          |           |
| **doubleAdder** |          **16** |   **4.828 ns** |  **5.7282 ns** |  **0.8880 ns** |   **0.04** |     **0.01** |    **7280 B** |
| interLocked |          16 | 122.182 ns |  4.9328 ns |  0.7647 ns |   1.00 |     0.00 |    7280 B |
|             |             |            |            |            |        |          |           |
| **doubleAdder** |          **32** |   **3.016 ns** |  **1.0519 ns** |  **0.1631 ns** |   **0.03** |     **0.00** |   **14064 B** |
| interLocked |          32 | 100.419 ns |  1.7495 ns |  0.2712 ns |   1.00 |     0.00 |   14064 B |

src/Logary/Metric/MetricRegistry.fs
```diff
@@ -0,0 +1,24 @@
+namespace Logary.Metric
+
+open System.Collections.Concurrent
+
+/// used for register metric, and can export metric infos
+type MetricRegistry() =
+  let metricBackStore = new ConcurrentDictionary<string, MetricExporter>()
```
haf: Do we need it to be concurrent? The configuration API is not intended to be called concurrently?

haf:
I had an idea behind the Reservoirs and the filter-callback/cep type APIs; to use it for metrics like this — each metric could then run on a "single thread" in the flow/cep area. What are your thoughts on this?

I'm looking ofrward to looking at how this all interacts with the rest of Logary.

src/Logary/Metric/MetricRegistry.fs
lust4life: it depend on user's usage, mertic can be register throuth a `LogManager`'s instance, or use a static way to use the default registry. each metric with the same name is singleton semantically. consider this semantically, it implements with thread safe.
usage: 
```f#
module A =

  let itemsInQueueGauge = GaugeConf.create "Items_In_Queue" "help info with metric name" |> LogManager.DefaultMetricRegistry.registerMetricWithNoLables

  let enqueue (items: #seq<_>) =
    let itemCount = items |> Seq.length |> float
    itemsInQueueGauge.inc itemCount
    // enqueue items logic
    

type B(logManager:LogManager) =
  let itemsInQueueGauge = GaugeConf.create "Items_In_Queue" "help info with metric name" |> logManager.metricRegistry.registerMetric

  member x.enqueue item =
    itemsInQueueGauge.noLabels.inc 1.
    // enqueu logic

type C() =
  let itemsInQueueGauge = GaugeConf.create "Items_In_Queue" "help info with metric name" |> LogManager.DefaultMetricRegistry.registerMetric

  member x.enqueue item =
    itemsInQueueGauge.noLabels.inc 1.
    // enqueu logic

```

if user use A style, then concurrent is no need, but B/C style usage need ensure same metric name get same metric.

lust4life:
> I had an idea behind the Reservoirs and the filter-callback/cep type APIs; to use it for metrics like this — each metric could then run on a "single thread" in the flow/cep area. What are your thoughts on this?

Yes, I started thinking like this, but find out it will have some pitfall during implementation.

- metric gauge/counter/histrogram are not like logs in a developer's view
  
  metric can occur more frequently, logs can be filter with level, if things in prod are ok, no one wants to check logs, so we use level/rules/pipes to filter things to send or send in a conditional way. But metric are usually needs to send for monitoring. so if we bind metric with logger/message, we need handle this situation. 
- since metric can occur more frequently and usually needs to be sent, it needs to have some light ways
  
  so it has performance limitation, if we use message as the data model, target will be the bottleneck when processing metric. since it run on a "single thread", if it can not catch up with metric real happens , then the monitoring/alert will be delay. 
  - https://github.com/Hopac/Hopac/issues/38  seems `Hopac` are optimized for low contention scenarios, so this maybe a potential problem
- prometheus are pull based
  
  - https://prometheus.io/blog/2016/07/23/pull-does-not-scale-or-does-it/#prometheus-is-not-an-event-based-system
  
  but target in logary seems more push-based, and more like a event-based-style, with prometheus, we just need grab these metrics when it pull these data, so target here seems a little unnatural. so the implementation right now is just some float value with its lables

Based on these ideas, a new data types(`metric`/`DoubleAdder`) are introduced in this implementation. what do you think ?

> I'm looking ofrward to looking at how this all interacts with the rest of Logary.

right now, the main api are mostly done. the rest is to 
- add a `Logary.Prometheus` project which use suave to listen a port from prometheus, grap metric data from `LogManager`'s `MetricRegistry`'s `getMetrictInfos`, then export to prometheus's text-based exposition format.
- some unit tests
- fix gap between `Message's Gauge` through pipe

haf:
> Yes, I started thinking like this, but find out it will have some pitfall during implementation.

> metric gauge/counter/histrogram are not like logs in a developer's view:
> metric can occur more frequently, logs can be filter with level, if things in prod are ok, no one wants to check logs, so we use level/rules/pipes to filter things to send or send in a conditional way. But metric are usually needs to send for monitoring. so if we bind metric with logger/message, we need handle this situation.

On one hand you're right; they may be used for different things.

On another; no — often there are logs that are also metrics/gauges with a singular count value. As such they are the same.

Another aspect is that an error is not alertable, but 50 errors the last second is. But you wouldn't know about it unless you have metrics, and then you'd like to show both the normal histogram of exception rates as well as the alertable histogram of exception rates; so you need the logs to be attached to the alert/sending to e.g. the OpsGenie target.

Another aspect still; you often have spans with child spans where each child span may contain a number of durations (gauges) of operations it performed, as well as the logs that were associated. As such, metrics + logs + tracing.

So I think pipes are great for collating this information together actually, since this information may be sent from many different threads/fibers.

That said, I'm not at all against your current direction of making the metric explicit; it's probably easier to start by making it explicit and then making it implicit/integrated once we know where we're going.

> since metric can occur more frequently and usually needs to be sent, it needs to have some light ways. so it has performance limitation, if we use message as the data model, target will be the bottleneck when processing metric. since it run on a "single thread", if it can not catch up with metric real happens , then the monitoring/alert will be delay.

I think you're assuming metrics are CPU bound when you say the target will be a contention point, but they are not; metrics is normally less than 1% of the CPU usage. If we're talking performance, I'd bet a better optimisation at this point would be to allow metrics to be updated without allocating a full Message object if and only if we know the metric is a metric only and not a log message.

But I hope that use-case can be an optimised version of the metric as Message in the long run.
      
> [Hopac/Hopac#38](https://github.com/Hopac/Hopac/issues/38)  seems `Hopac` are optimized for low contention scenarios, so this maybe a potential problem

There'd be no contention as in the many-writer-threads scenario, we have a fan-in pattern and the metric/reservoir can work at-CPU-speed with batched inputs if we go the metric-as-message route (perhaps with a singleton message per registered w/ updated timestamp), which should give good cache locality.

> prometheus are pull based
> but target in logary seems more push-based, and more like a event-based-style, with prometheus, we just need grab these metrics when it pull these data, so target here seems a little unnatural.  the implementation right now is just some float value with its lables

I agree that Target may not be the right level of abstraction, but synchronous commit type concurrency internally for events and polling externally, is IMO a perfectly good abstraction. Here's a health-check type API I have for an internal project at qv2, that implements a pollable API for Prometheus type health/metric checks:

```fsharp

open Hopac
open Hopac.Infixes
open ReCQ.Logging
open ReCQ.Logging.Message
open ReCQ.StreamProcessing
open NodaTime

type NodeRef =
  private {
    spnId_: SPNId
    reader: Reader
    writer: Writer
    shutdownCh: Ch<IVar<unit>>
  }

  member x.spnId = x.spnId_

  static member create spnId r w shC =
    { spnId_ = spnId; reader = r; writer = w; shutdownCh = shC }

module NodeRef =
  let private logger = Log.create "Qvitoo.SPNs.NodeRef"

  let create (es: EventStoreImpl.EventStore, infra: SPNInfrastructure, clock: IClock)
             (node: Node<_,_>): Job<NodeRef> =
    let shutdown = Ch ()
    let srw = infra.reader.bindToContext (InternalSerPair.create node.serialiser node.deserialiser)

    let startWriter =
      { Writer.WriterConfig.wall = clock
        Writer.node = node
        Writer.srw = srw
        Writer.infra = infra.writer }
      |> Writer.start

    let startReader =
      { Reader.ReaderConfig.wall = clock
        Reader.node = node
        Reader.srw = srw
        Reader.infra = infra.reader }
      |> Reader.start

    let waitForShutdown reader writer =
      shutdown ^=> fun reply ->
        Writer.shutdown writer <+> Reader.shutdown reader
        >>=. IVar.fill reply ()

    (startReader <&> startWriter) >>= fun (reader, writer) ->
    Job.start (waitForShutdown reader writer)
    >>-. NodeRef.create node.spnId reader writer shutdown

  let shutdown (n: NodeRef): Alt<unit> =
    n.shutdownCh *<+=>- id

  let getStatus (n: NodeRef): Alt<SPNId * ReaderStatus * WriterStatus> =
    // consider lifting into the loop inside the 'create' function.
    (Reader.getStatus n.reader <+> Writer.getStatus n.writer) ^-> fun (rS, wS) ->
    n.spnId, rS, wS

type SPNController =
  private {
    registerNodeCh: Ch<NodeRef * IVar<unit>>
    getStatusCh: Ch<IVar<Map<SPNId, ReaderStatus * WriterStatus>>>
    shutdownCh: Ch<IVar<unit>>
    es: EventStoreImpl.EventStore
    infra: SPNInfrastructure
    clock: IClock
  }
  static member internal create nC stC sdC es infra clock =
    { registerNodeCh = nC; getStatusCh = stC; shutdownCh = sdC
      es = es; infra = infra; clock = clock }

module SPNController =

  let private logger = Log.create "Qvitoo.SPNs.SPNController"

  type State =
    { health: Map<SPNId, ReaderStatus * WriterStatus>
      nodes: NodeRef list }

    static member create () =
      { health = Map.empty; nodes = [] }

  let create (connection: ReCQ.EventStore.Connection, wallClock: IClock option): Job<SPNController> =
    let wallClock = wallClock |> Option.defaultValue (SystemClock.Instance :> IClock)
    let registerNode, getStatus, statusTick, shutdown = Ch (), Ch (), Ch (), Ch ()

    let ticker =
      Job.foreverServer (timeOut (System.TimeSpan.FromSeconds 5.) ^=> Ch.give statusTick)

    let statusFolder (health: Map<SPNId, _>) (spnId, rS, wS) = health |> Map.add spnId (rS, wS)
    let setHealth (state: State) (health: Map<_,_>) = { state with health = health }

    let pollEveryNode (state: State) (): Job<State> =
      Job.conCollect (state.nodes |> List.map NodeRef.getStatus) >>- fun statuses ->
      statuses
        |> Seq.fold statusFolder state.health
        |> setHealth state

    let getStatusRequest state (reply: IVar<Map<SPNId, _>>): Job<_> =
      reply *<= state.health

    let registerNodeRequest state (nodeRef: NodeRef, reply: IVar<unit>): Job<State> =
      let nextState = { state with nodes = nodeRef :: state.nodes }
      IVar.fill reply () >>-. nextState

    let shutdownRequest state (reply: IVar<unit>): Job<State> =
      Job.conIgnore (state.nodes |> List.map NodeRef.shutdown)
      >>= IVar.fill reply
      >>-. state

    let initialise =
      // Starts a background job that triggers node state polling every X seconds.
      ticker
      // The EventStoreImpl adapts the ReCQ.EventStore.Connection interface.
      // There's also an implementation available for an in memory version of this.
      >>=. EventStoreImpl.EventStore.create connection >>- fun es ->
      // There's also `let esPure = EventStoreImpl.EventStore.createPure connection` available to speed up writes
      es, EventStoreImpl.EventStore.toInterface es

    let iter (state: State) =
      Alt.choosy [|
        getStatus ^=> (getStatusRequest state >->. state)
        statusTick ^=> pollEveryNode state
        registerNode ^=> registerNodeRequest state
        shutdown ^=> shutdownRequest state
      |]

    initialise >>= fun (es, infra) ->
    Job.iterateServer (State.create ()) iter
    >>-. SPNController.create registerNode getStatus shutdown es infra wallClock

  let registerNode (c: SPNController) (n: Node<_,_>): Alt<unit> =
    c.registerNodeCh *<-=>= fun reply ->
    NodeRef.create (c.es, c.infra, c.clock) n >>- fun nodeRef ->
    nodeRef, reply

  let shutdown (c: SPNController): Alt<unit> =
    c.shutdownCh *<+=>- id

  let getStatus (c: SPNController): Alt<Map<SPNId, ReaderStatus * WriterStatus>> =
    c.getStatusCh *<+=>- id
```
// API:

```fsharp

  let getHealth spnCP: WebPart =
    fun ctx ->
      async {
        let! status =
          if Promise.Now.isFulfilled spnCP then
            let spnC = Promise.Now.get spnCP
            let status =
              Alt.choose [
                SPNController.getStatus spnC
                timeOutMillis 100 ^-> fun () ->
                  logger.info (eventX "Timeout talking to SPN Controller")
                  Map.empty
              ]
            Job.toAsync status
          else
            Async.result Map.empty

        return! OK (buildStatusFrom status) ctx
      }
```

> Based on these ideas, a new data types(`metric`/`DoubleAdder`) are introduced in this implementation. what do you think ?

I absolutely think you should follow your gut to start with and then we'll discuss your design and see if we can find improvements to it!

lust4life:
thx！will reply it later（off the laptop）, nice review.

lust4life:
> On another; no — often there are logs that are also metrics/gauges with a singular count value. As such they are the same.
> Another aspect is that an error is not alertable, but 50 errors the last second is. But you wouldn't know about it unless you have metrics, and then you'd like to show both the normal histogram of exception rates as well as the alertable histogram of exception rates; so you need the logs to be attached to the alert/sending

about this part, we can `enableHookMetric` on logary, which will treat `span` as prometheus's gauge, send logary's gauge to prometheus's gauge, treat log message as prometheus's counter if message use `enableCounterMetric` like below : 
```
Message.event Error "error message" 
|> Message.enableCounterMetric  {name = "some_method_error_count"; description="description"; labelNames=[||]}
|> logger.xxxx
```
through this, we can hook gauge/log in logary with metric, what do you think ? and will check code you paste above later and then give more feedback.

src/Logary/MessageModule.fs
```diff
@@ -186,14 +186,41 @@ module Message =
...
+  [<CompiledName "AddGauges">]
+  let addGauges (gauges: #seq<string * Gauge>) message =
+    (message, gauges) ||> Seq.fold (fun m (n, g) -> m |> addGauge n g)
+
+  /// if one log message needs metric its count, then we can invoke this method for shortcut.
+  /// if one needs metric gauges or histogram, it's better to define a metric type to handle this case.
+  /// since these metric are not bound to message/logger, and can be cached by filed declaration (more efficient).
+  /// labelNames will grab its value from message's context: fields -> contexts, if not found, will raise error
+  [<CompiledName "EnableCounterMetric">]
+  let enableCounterMetric (counterConf: Metric.BasicConf) (message: Message) =
```
haf: hmmmmmm....

I don't like this design TBH... Do we really have to store it in the Message...?

src/Logary/Registry.fs
```diff
@@ -207,6 +214,60 @@ module Registry =
...
+              )
+            counterMetric.labels labelValues
+
+        counter.inc 1.
+
+
+      // process logary's gauge
+      let sensorName = message.name |> PointName.format
+      let gauges = message |> Message.getAllGauges
+      gauges |> Seq.iter (fun (gaugeName, gauge) ->
```
haf: Can't we just have the metric instance match towards the point name instead of the above "counter metric"?

lust4life: here is for hook up logary's `Gauge` with prometheus's gauge, legacy code using  `gaugeWithUnit` in `MessageModule` api, can get this benefit by one configuration as logary's initiation.

src/Logary/MessageModule.fs
lust4life:
```f#

module A =
  open Logary.Metric

  let logger = Log.create "xxx.yyy"

  let parseConfigFailedCounter =
    GaugeConf.create {name="xxx_yyy_parse_config_faild_total"; description= "the total number for xxx.yyy.parseConfig faild"; labelNames = [| "path" |]}
    |> LogManager.DefaultMetricRegistry.registerMetric

  // add gauge metric and histogram with buckets [| 0.001; 0.005; 0.01; 0.025; 0.05; 0.1; 1.0|] in the same time
  let parseConfigGaugeWithHistogram =
    GaugeConf.create {name="xxx_yyy_parse_config_latency_seconds"; description= "xxx.yyy.parseConfig latency in seconds"; labelNames = [||]}
    |> GaugeConf.withHistogram [| 0.001; 0.005; 0.01; 0.025; 0.05; 0.1; 1.0|]
    |> LogManager.DefaultMetricRegistry.registerMetric
    |> Metric.noLabels


  // using default buckets:  [| 0.005; 0.01; 0.025; 0.05; 0.075; 0.1; 0.25; 0.5; 0.75; 1.; 2.5; 5.0; 7.5; 10. |]
  let parseConfigLatencyHistogram =
    HistogramConf.create("xxx_yyy_parse_config_latency_seconds", "xxx.yyy.parseConfig latency in seconds")
    |> LogManager.DefaultMetricRegistry.registerMetric
    |> Metric.noLabels

  // using hookup and shortcut style
  let parseConfigWithShortcut path =
    try
      // if we hook up metric at logary's conf, then there is no need to do other things here.
      // span will be treated as gauges when span has been finished.
      use traceSpan =
        logger.buildSpan()
        |> Span.setMessage (eventX "time usage about parse config with {path}" >> setField "path" path)
        |> Span.start

      // do some work here
      failwith "not implement"
      
    with ex ->
      eventX "parse config from {path} failed"
      >> setField "path" path
      >> enableCounterMetric {name="xxx_yyy_parse_config_faild_total"; description= "the total number for xxx.yyy.parseConfig faild"; labelNames = [| "path" |]}
      |> logger.error

  // using metric as a separate instance style
  let parseConfig path =
    try
      let traceSpan =
        logger.buildSpan()
        |> Span.setMessage (eventX "time usage about parse config with {path}" >> setField "path" path)
        |> Span.start

      // do some work here
      failwith "not implement"


      let spanLog = traceSpan.finish id
      // gauge with histogram like below
      parseConfigGaugeWithHistogram.set (Duration.FromTicks(spanLog.duration).TotalSeconds)
      // or metric only histogram like below
      parseConfigLatencyHistogram.observe (Duration.FromTicks(spanLog.duration).TotalSeconds)

    with ex ->
      eventX "parse config from {path} failed"
      >> setField "path" path
      |> logger.error

      (parseConfigFailedCounter.labels [| path |]).inc 1.

```

`enableCounterMetric` is just some shortcut for creating a counter metric for a log message (means its occurrence total count, if this message is not send because its rules or logging level's config, this counter will not metric either), we can use `metric` separately like code above `parseConfig` . We recommend that use metric separately, even if the message are not send, the counter still metric. and using metric separately will not have to do a find operation to get the real metric, can save some time. 

> These should be primarily used as file-static variables, that is, global variables defined in the same file as the code they’re instrumenting. The client library SHOULD enable this. The common use case is instrumenting a piece of code overall, not a piece of code in the context of one instance of an object. Users shouldn’t have to worry about plumbing their metrics throughout their code, the client library should do that for them (and if it doesn’t, users will write a wrapper around the library to make it "easier" - which rarely tends to go well).

says from https://prometheus.io/docs/instrumenting/writing_clientlibs/#metrics

lust4life: or do you have any ideas how we can design a api to wrapper these inside ?
prometheus's metric needs:
- metric name 
- metric help (optional)
- labels (`Map<string,string>`) (optional, depend on use case)

each metric with same labels needs use the same metric, 
```f#
let metric = Metric("metricA").withLabelName("http_verb")
metric.label("get").inc 5
metric.label("get").inc ()
metric.label("post").inc ()
```
code above need export metric as:
metricA{http_verb=get} 6
metricA{http_verb=post} 1

src/Logary/Registry.fs
lust4life: since message know nothing about the registry/logger, but metric needs register on some `MetricRegistry` which in a logary's `LogManager/Registry`, so here we need store info in message, when dispatch by registry, grab from it and hook with `metricRegistry`

src/Logary/MessageModule.fs
haf: I see your points, but I don't think that what you've described is an optimal solution. If you commit that file in here, I can see if I can make the API more to my liking in this PR

haf: `gauge "metric_a" 6 |> setField "http_verb" "get"`

haf: In your exporter then; `exporter |> varyBy "http_verb"`

haf: or even default the exporter to vary always for every Map of field it sees.

src/Logary/Extensions/LogManager.fs
```diff
@@ -0,0 +1,15 @@
+namespace Logary
+
+open Logary.Metric
+
+open System.Runtime.CompilerServices
+
+[<AutoOpen; Extension>]
+module LogManagerEx =
+  let defaultMetricRegistry = new MetricRegistry()
```
haf: I strongly feel this should go the way of how we initialise the other registry and have a pointer it from here; a projection from the singleton of the last-initialised logmanager. I've updated https://github.com/haf/expecto/blob/master/Expecto/Logging.fs#L366-L389 to do that; and I would like to replace our current clock+flyweights to use the way we do it in Expecto.

src/Logary/Global.fs
```diff
@@ -95,6 +89,7 @@ module internal Global =
   let lockSem fn =
     lock (getConsoleSemaphore ()) fn
```
haf: DVars here, that is

haf:
If we add metrics explicitly, it's IMO important we can have them output as message as well as in an exporter, otherwise, we can't send to influx or pass them between rutta and other rutta or other targets.

haf:
BTW I think most of this PR is looking really good! Keep up the good work!

src/Logary/Global.fs
lust4life: nice, will try it later

src/Logary/Extensions/LogManager.fs
lust4life: `MetricRegistry` seems a litter bit different, even if people change logary's global registry by `setGlobals`, the `defaultMetricRegistry` should not changed by that registry's metricRegisty. That is why it use a `static member DefaultMetricRegistry = ...`, log message or metrics can occur before logary's registry‘s initialise. log message can be dropped that's why we have a nulllogger, but since metrics is lighter, we should preserve these metrics before registry's initialise. 

If we use `DVar` here, we need do some merges (the old metrics and the new metrics), which seems not easy, especially when user using these metric at same time.

lust4life: but you are right, it will be a good replace for our current clock+flyweights loggers. will change it in another branch

lust4life:
@haf  add tests. and the main usage can be found in these tests. test with prometheus in local machine, it can export text protocol to prometheus and query with grafana.

haf:
Nice! I have to dive into your changes and review them and play around with them. Without first having done that, does what you have implemented now support this use case?

We have IoT metrics being fed into Rutta, three instances of them. Each message has a `value`, that corresponds to a "message type" in a regular messaging system. Some messages are interesting; we'd like to create a histogram based on e.g. message type "scanned QR"-valued messages' field: "vehicle state", which can then be fed as Message-values into e.g. Google PubSub. I would have liked to write it with the Flow feature, but I suppose that it would in part intersect with your new histograms here. How would you do it?

Also, would you do it differently when Rutta is replicated x3 or even x10, in order to handle the load? The histogram would be similar but slightly different between the replicas, which could be just fine.

src/Logary.PerfTests/DoubleAdder.fs
```diff
@@ -0,0 +1,133 @@
...
+
+    [<Benchmark>]
+    member x.doubleAdder () = doubleAdderCounter.Add 2.
+
+  type Counter() =
+    let mutable interlockedCounter = 0L
+
+    let interLockedAdd value =
+      let mutable loop = true
+      while(loop) do
```
haf: spaces

src/Logary.PerfTests/Program.fs
```diff
@@ -144,7 +144,7 @@ module Tests =
           jobs = [ xJ ]
       }
 
-    testList "benchmarks" [
+    ptestList "benchmarks" [
```
haf: remove `p`?

src/Logary.Prometheus.Tests/Program.fs
```diff
@@ -0,0 +1,122 @@
...
+      use cancelWebServer =  { new IDisposable with member x.Dispose() = cts.Cancel () }
+
+      ExporterConf.create "/metrics" LogManager.DefaultMetricRegistry
+      |> Exporter.runAsync cts.Token
+      |> ignore
+
+      let gauge =
+        GaugeConf.create {name = "time_latancy"; description = "test time latancy"; labelNames = [| "endpoint" |]}
+        |> GaugeConf.withHistogram [|50.; 100.; 500.;|]
+        |> LogManager.DefaultMetricRegistry.registerMetric
```
haf: Do you think we could write this test on a logary instance? I'd prefer to have test non-interfering with each other. And this is a global, right?

src/Logary.Prometheus.Tests/Program.fs
```diff
@@ -0,0 +1,122 @@
...
+  let checkTextProtocol expected actual =
+    expected |> String.split '\n' |> Expect.containsAll "should exported to prometheus text protocol" (actual |> String.trimStart |> String.split '\n')
+
+
+  [<Tests>]
+  let tests = testList "Logary.Prometheus" [
+
+
+    testCaseJob "export to prometheus text protocol" <| job {
+      let httpClient = new HttpClient()
```
haf: Extract from test into its own function

src/Logary.Prometheus.Tests/Program.fs
```diff
@@ -0,0 +1,122 @@
...
+      let! exportedText = exportedTextJob
+
+      checkTextProtocol expected exportedText
+
+    }
+
+
+
+
+    testCase "dfault metric name transformer" <| fun () ->
```
haf: spelling

src/Logary.Prometheus.Tests/Program.fs
```diff
@@ -0,0 +1,122 @@
...
+      ExporterConf.defaultMetricNameTrans "items in queue" |> Expect.equal "should transform" "items_in_queue"
```
haf: Nice

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+open Suave.Operators
+open Suave.Writers
+open Hopac
+
+
+type ExporterConf =
+  {
+    metricNameTransformer: string -> string
+    metricRegistry: MetricRegistry
+    webConfig: SuaveConfig
```
haf: It feels a bit like this should be under a namespace `Logary.Prometheus.Exporter` or something that relates to the polling, so that the first types in the `.Prometheus` namespace aren't polled-exporters. Thinking that we might want to feed prometheus with push at some point, e.g. when we would like to ingest historical data from Logary.

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+  let defaultMetricNameTrans metricName =
+    let mutable nameResult = metricName
+    if String.IsNullOrEmpty nameResult then failwith "name can't be null"
+    if reservedNameRegex.IsMatch nameResult then
+      nameResult <- "metric_" + nameResult.Substring 2
+    if not <| startCharValidRegex.IsMatch nameResult then
+      nameResult <- "metric_" + nameResult
+    notValidAfterPartRegex.Replace(nameResult, "_")
+
+  let nameValidator name =
```
haf: Could this validator be named `validateName` return `Result.Ok`/`Result.Error` and then have another, `forceValidateName` that does `Result.getOrThrow`? Just to keep the default case total and the "easy" case throwing; aiming to have composable APIs.

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+      metricRegistry = registry
+      webConfig = defaultConfig
+      urlPath =  urlPath
+    }
+
+  let webConfig webConfig conf =
+    { conf with webConfig = webConfig }
+
+
+module Exporter =
```
haf:
```
namespace Logary.Prometheus.Exporter

module Formatting = ...
```
?

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+  open System.Threading
+  open System.Globalization
+  open Logary.Internals.Chiron.Formatting
+  open System.IO
+
+  /// HELP lines may contain any sequence of UTF-8 characters (after the metric name),
+  /// but the backslash and the line feed characters have to be escaped as \\ and \n, respectively
+  let escapeHelp (help: string) =
+    help.Replace("\\", @"\\").Replace("\n",@"\n")
```
haf: I'm not going to argue here at all, but we should consider using `netcoreapp2.2` in the future, and defaulting to the non-allocating Span-based API:s.

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+  let formatFloat (value: float) =
+    if Double.IsPositiveInfinity value then "+Inf"
+    elif Double.IsNegativeInfinity value then "-Inf"
+    elif Double.IsNaN value then "Nan"
+    else value.ToString(CultureInfo.InvariantCulture)
+
+  let appendLine (sb: StringBuilder) s =
+    sb.AppendLine s |> ignore
+
+  let getTypeStr (metricInfo: MetricInfo) =
```
haf: Nicer to have on the `MetricInfo` type, IMO

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+    match metricInfo with
+    | Gauge _ -> "gauge"
+    | Histogram _ -> "histogram"
+
+  let getIdentifier nameTransform (name: string) (labels: Map<string,string>) =
```
haf: Is it intended not to use Logary's `HashMap`? I know it's harder to use though. Same comment here as the formatting above about Span-based APIs - not required to fix though.

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+  //  http_request_duration_seconds_bucket{le="0.5"} 129389
+  //  http_request_duration_seconds_bucket{le="1"} 133988
+  //  http_request_duration_seconds_bucket{le="+Inf"} 144320
+  //  http_request_duration_seconds_sum 53423
+  //  http_request_duration_seconds_count 144320
+  let exportMetricInfo (sb: StringBuilder) nameTransform (metricInfo: BasicInfo * seq<MetricInfo>) =
+    let basicInfo, metricInfos = metricInfo
+    let metricName = nameTransform basicInfo.name
+    let help = escapeLabelValue basicInfo.description
+    let metricType = metricInfos |> Seq.first |> Option.map getTypeStr
```
haf: This is double iteration of the seq. if this API is primarily internal for implementing functionality, consider using a concrete data-type that has O(1) indexing functionality, like an array.

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+    if metricType.IsSome then
+      appendLine sb (sprintf "# HELP %s %s" metricName help)
+      appendLine sb (sprintf "# TYPE %s %s" metricName metricType.Value)
+      for metricInfo in metricInfos do
```
haf: extract iteration body to function

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+    let metricNameTransformer = exportConf.metricNameTransformer >> ExporterConf.nameValidator
+    let metricInfos = metricRegistry.getMetrictInfos ()
+    let sb = new System.Text.StringBuilder ()
+    metricInfos |> Seq.iter (exportMetricInfo sb metricNameTransformer)
+    sb.ToString()
+
+
+  let internal exportWebPart exportConf : WebPart =
+    let urlPath = exportConf.urlPath
+    path urlPath >=> GET  >=> setHeader "Content-Type" "text/plain; charset=utf-8" >=> request (fun r ->  Successful.OK (exportToPrometheus exportConf))
```
haf: Nice, but perhaps a few line breaks?

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+
+
+  let runAsync token conf  =
```
haf: Why do we run it as a Task? Also, start as Task doesn't match up with the name of this function.

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+    member x.set value =
+      gaugeValue <- new DoubleAdder(value)
+      histogram |> Option.iter(fun histogram -> histogram.observe value)
+
+    member x.explore () =
+      let confInfo = conf.basicInfo
+      let basicInfo = { name= confInfo.name; description = confInfo.description }
+      let gaugeValue =  gaugeValue.Sum()
+      let metricInfo = { labels = labels; gaugeValue = gaugeValue }
+      (basicInfo, MetricInfo.Gauge metricInfo)
```
haf: Unnecessary parenthesis

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+  let escapeHelp (help: string) =
+    help.Replace("\\", @"\\").Replace("\n",@"\n")
+
+  /// label_value can be any sequence of UTF-8 characters,
+  /// but the backslash (\), double-quote ("), and line feed (\n) characters
+  /// have to be escaped as \\, \", and \n, respectively.
+  let escapeLabelValue (labelValue: string) =
+    labelValue.Replace("\\", @"\\").Replace("\"",@"\""").Replace("\n",@"\n")
+
+  let formatFloat (value: float) =
```
haf: Nice

src/Logary.Prometheus/Exporter.fs
```diff
@@ -0,0 +1,171 @@
...
+    if Double.IsPositiveInfinity value then "+Inf"
+    elif Double.IsNegativeInfinity value then "-Inf"
+    elif Double.IsNaN value then "Nan"
+    else value.ToString(CultureInfo.InvariantCulture)
```
haf: Nice. We also have `Culture.invariant` available.

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+let private gaugeConf = GaugeConf.create basicConf
+
+let tests = [
+
+  testCase "register metric" <| fun () ->
+    let registry = new MetricRegistry()
+    let gauge = gaugeConf |> registry.registerMetric
+    gauge.noLabels.inc 5.
+    let basicInfo, MetricInfo.Gauge gaugeInfo = gauge.noLabels.explore ()
+    basicInfo |> Expect.equal "should have same basic info" { name = gaugeConf.basicInfo.name; description = gaugeConf.basicInfo.description }
```
haf: Rather test the two properties on two lines

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+    let gauge = gaugeConf |> registry.registerMetric
+    gauge.noLabels.inc 15.
+    let gauge = gaugeConf |> registry.registerMetric
+    gauge.noLabels.dec 10.
+
+    let _, MetricInfo.Gauge gaugeInfo = gauge.noLabels.explore ()
+    gaugeInfo.gaugeValue |> Expect.equal "should reuse the same one" 5.
+
+    Expect.throws "should throws when register metric with same name but different basic conf " <| fun () ->
+      GaugeConf.create {basicConf with labelNames = [| "with some label" |]} |> registry.registerMetric |> ignore
```
haf: Is there a similar `tryCreate` function that doesn't throw? Same thing here about composable API:s... Don't want users to have to add try-catch if they interact with Logary, ever, unless they opt into it.

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+    let gauge = gaugeConf |> registry.registerMetric
+    gauge.noLabels.inc 10.
+    let _, MetricInfo.Gauge gaugeInfo = gauge.noLabels.explore ()
+    gaugeInfo.gaugeValue |> Expect.equal "registerMetricWithNoLables is just a shotcut for registerMetric(gaugeConf).noLabels" 20.
+
+
+  testCase "register metric with labels" <| fun () ->
+
+    Expect.throws "should throws when gauge with label value but no label names set up" <| fun () ->
+        let registry = new MetricRegistry()
```
haf: 2 space indent here

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+  let mutable gaugeValue = new DoubleAdder()
+  let changeGaugeValueLock = obj ()
+
+  let tryDoHistogramObserve () =
+    histogram |> Option.iter(fun histogram ->
+      let sum = gaugeValue.Sum ()
+      histogram.observe sum)
+
+  interface IGauge with
+    member x.inc value =
```
haf: This is a counter, not a gauge? https://godoc.org/github.com/prometheus/client_golang/prometheus#hdr-A_Basic_Example Gauges only have `set (value: float)`

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+      | Some histogram ->
+        lock changeGaugeValueLock <| fun () ->
+          gaugeValue.Add value
+          gaugeValue.Sum () |> histogram.observe
+      | None -> gaugeValue.Add value
+
+    member x.dec value =
+      (x :> IGauge).inc -value
+
+    member x.set value =
```
haf: Should be decouple gauge from counter, perhaps?

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+    member x.explore () =
+      let confInfo = conf.basicInfo
+      let basicInfo = { name= confInfo.name; description = confInfo.description }
+      let gaugeValue =  gaugeValue.Sum()
+      let metricInfo = { labels = labels; gaugeValue = gaugeValue }
+      (basicInfo, MetricInfo.Gauge metricInfo)
+
+module GaugeConf =
+
+  /// Use this method carefully, since this will calculate each gague's change value as the ovserve data for histogram.
```
haf: gague -> gauge

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+  /// And it will cause a lock to change gauge value.
+  /// Consider using histogram seperately
+  let enableHistogram histogramConf gaugeConf =
+    { gaugeConf with histogramConf = Some histogramConf }
+
+  /// Use this method carefully, since this will calculate each gague's change value as the ovserve data for histogram.
```
haf: gague -> gauge

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
...
+  /// And it will cause a lock to change gauge value.
```
haf: Could you clarify this comment?

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
+namespace Logary.Metric
+open System
+
+type HistogramConf=
+  {
+    basicInfo: BasicConf
+    buckets: float array
```
haf: use `float[]` syntax plx

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
+namespace Logary.Metric
+open System
+
+type HistogramConf=
```
haf: space

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
...
+  {
+    basicInfo: BasicConf
+    buckets: float array
+  }
+
+  interface MetricBuilder<IHistogram> with
+    member x.basicConf = x.basicInfo
+    member x.build lables registry = new Histogram(x, lables) :> IHistogram
```
haf: labels

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
...
+
+    member x.explore () =
+      let confInfo = conf.basicInfo
+      let basicInfo = { name= confInfo.name; description = confInfo.description }
+      let sumInfo =  sumCounter.Sum()
+      let bucketsInfo = sortedBucketCounters |> Map.map (fun upperBound counter -> counter.Sum())
+      let metricInfo = { labels = labels; sumInfo = sumInfo; bucketsInfo = bucketsInfo }
+      (basicInfo, MetricInfo.Histogram metricInfo)
+
+module HistogramConf =
```
haf: Nice

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
...
+
+  let defaultBuckets = [| 0.005; 0.01; 0.025; 0.05; 0.075; 0.1; 0.25; 0.5; 0.75; 1.; 2.5; 5.0; 7.5; 10. |]
```
haf: Why these values? Comment in the code?

src/Logary/Metric/Histogram.fs
```diff
@@ -0,0 +1,79 @@
...
+
+  let buckets buckets conf =
+    { conf with buckets = buckets }
+
+  let linearBuckets start width count conf =
+    if count < 1 then failwith "count needs positive"
```
haf: `invalidArg`

src/Logary/Registry.fs
```diff
@@ -58,6 +62,9 @@ module Registry =
 
       /// default logger rules from logary conf, will be applied when create new logger
       defaultLoggerLevelRules: (string * LogLevel) list
+
+      /// metric registry
+      metricRegistry: MetricRegistry
```
haf: Nice

src/Logary/Registry.fs
```diff
@@ -206,6 +214,59 @@ module Registry =
          && not ("true" = Env.varDefault "LOGARY_I_PROMISE_I_HAVE_PURCHASED_LICENSE" (fun () -> "false")) then
          failwith "You must purchase a license for Logary to run it on or with IIS or Kestrel."
 
+    // check if msg has enable metric counter
+    let hookupMetric (metricRegistry: MetricRegistry) (message: Message) =
```
haf: This is what I was thinking could be a default `Flow` middleware.

src/Logary/Registry.fs
```diff
@@ -206,6 +214,59 @@ module Registry =
...
+          | labelNames ->
+            // find lable name's value from message's context, fields goes first
+            let labelValues =
+              labelNames |> Array.map (fun labelName ->
+                match message |> Message.tryGetField labelName with
+                | Some labelValue -> string labelValue
+                | None ->
+                  match message |> Message.tryGetContext labelName with
+                  | Some labelValue -> string labelValue
+                  | None -> failwithf "label name (%s) not found in message,maybe you should metric with nolables" labelName
```
haf: Do we want to throw from within Logary? :-s Also; spelling.

src/Logary/Registry.fs
```diff
@@ -206,6 +214,59 @@ module Registry =
...
+        let gaugeMetric =
+          GaugeConf.create { name = sensorName; description = sensorName; labelNames = [| "gauge_name" |]}
+          |> metricRegistry.registerMetric
+          |> Metric.labels [| gaugeName |]
+
+        let gaugeValue = gauge.value.toFloat ()
+        gaugeMetric.set gaugeValue
+      )
+
+      // process logary's span as gauge
```
haf: Really nice!

src/ingestion/Logary.Ingestion.HTTP/AssemblyInfo.fs
```diff
@@ -2,17 +2,17 @@
 namespace System
 open System.Reflection
 
-[<assembly: AssemblyTitleAttribute("Logary.Ingestion.HTTP")>]
-[<assembly: AssemblyProductAttribute("Logary.Ingestion.HTTP")>]
+[<assembly: AssemblyTitleAttribute("Logary.Ingestion.Http")>]
```
haf: Do you need to `rm -rf Logary.Ingestion.Http` and then `git checkout -- Logary.Ingestion.HTTP` to get the folder name correct, perhaps?

src/targets/Logary.Targets.Elasticsearch/AssemblyInfo.fs
```diff
@@ -2,17 +2,17 @@
 namespace System
 open System.Reflection
 
-[<assembly: AssemblyTitleAttribute("Logary.Targets.Elasticsearch")>]
-[<assembly: AssemblyProductAttribute("Logary.Targets.Elasticsearch")>]
+[<assembly: AssemblyTitleAttribute("Logary.Targets.ElasticSearch")>]
```
haf: Same here

lust4life: thx, finally know why this happened

src/ingestion/Logary.Ingestion.HTTP/AssemblyInfo.fs
lust4life: it works, finally know why this happened

src/Logary/Registry.fs
lust4life: since if pipe doesn't introduce any async/parall processing, this will fail at user's callsite, but given a warning message to its label value may be a good idea. let user find this issue when query metric at backend (like prometheus)

src/Logary/Registry.fs
lust4life: Yes, make it as a middleware.

src/Logary.Prometheus.Tests/Program.fs
lust4life: Yes, you are right. fixed.

src/Logary.Prometheus/Exporter.fs
lust4life: 👍

src/Logary.Prometheus/Exporter.fs
lust4life: 😺

src/Logary.Prometheus/Exporter.fs
lust4life: get it

src/Logary.Tests/Metrics.fs
lust4life: fixed, BTW, do you know is there any auto formatting tools for fsharp ?

src/Logary.Prometheus/Exporter.fs
lust4life: you are right, 👍

lust4life:
thanks for the detailed review  ❤️  , will reply remains review later.

src/Logary.Prometheus/Exporter.fs
lust4life: done.

src/Logary.Prometheus/Exporter.fs
lust4life: what do you mean ❓

src/Logary.Prometheus/Exporter.fs
lust4life: labels are used in two cases, 
- when register a labeled metric, it will use labels(name-value paris) as the key of the metric, `let metricStore = new ConcurrentDictionary<Map<string,string>, 't>()`, `Map` can give the structural equality which `HashMap` doesn't.  no more lookups here.
- when export as metrc info to backend
  in this case, it will just loop for formatting, seems `HashMap` here give less advantage here.

src/Logary.Prometheus/Exporter.fs
lust4life: Added a issue for tracking

src/Logary.Tests/Metrics.fs
lust4life: this is intentional, according to this : https://prometheus.io/docs/instrumenting/writing_clientlibs/#labels
> Client libraries MUST NOT under any circumstances allow users to have different label names for the same metric for Gauge/Counter/Summary/Histogram or any other Collector offered by the library.

fail fast will force user to check their metric name definition.

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+    let registry = new MetricRegistry()
+    let gauge = {basicConf with labelNames = [| "queue name" |]} |> GaugeConf.create |> registry.registerMetric
+    (gauge.labels [| "some queue name" |]).set 10.
+    gauge.noLabels.set 20.
+
+    let metricInfos = registry.getMetrictInfos ()
+    metricInfos |> Seq.length |> Expect.equal "should have only one metric" 1
+
+    let _, metricInfoDetails = metricInfos |> Seq.head
+    metricInfoDetails |> Seq.length |> Expect.equal "should have two metric info, one with label name, one with no labels" 2
```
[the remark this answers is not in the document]
lust4life: seq here is for lazy load. if user want to avoid a re-iterate, they can cast them to a array/list first.

src/Logary/Metric/Histogram.fs
lust4life: https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L56-L61 from here 😸   Comment added.

src/Logary/Metric/Gauge.fs
lust4life:
``` 
  /// Consider using histogram seperately.
  /// Technically, a gauge can also be a histogram, so if you want to measure some histograms when gauge something,
  /// you can use this method. However, because the gauge can use the inc/dec method compared to the histogram,
  /// if you need to use the histogram when gauge, it will cause the sum operation after inc/dec.
  /// In order to ensure the histogram is more accurate under multi-threading, the lock will be performed.
  /// But if it is a simple gauge, which does not use lock for multithreaded inc/dec
```
clarify this comment

src/Logary/Metric/Gauge.fs
lust4life: https://godoc.org/github.com/prometheus/client_golang/prometheus#Gauge

>  Gauge is a Metric that represents a single numerical value that can arbitrarily go up and down.
>
> A Gauge is typically used for measured values like temperatures or current memory usage, but also "counts" that can go up and down, like the number of running goroutines.

src/Logary/Metric/Gauge.fs
lust4life: this is what i was wondering,  according to this:

- https://prometheus.io/docs/instrumenting/writing_clientlibs/#counter
- https://godoc.org/github.com/prometheus/client_golang/prometheus#Counter

> Counter is a Metric that represents a single numerical value that only ever goes up. That implies that it cannot be used to count items whose number can also go down, e.g. the number of currently running goroutines. Those "counters" are represented by Gauges.

>A Counter is typically used to count requests served, tasks completed, errors occurred, etc. 

i don't really know if a `counter` is different with `gauge` at prometheus's under implementation, or prometheus treat them the same thing.

if they are only different at the semantic level, i think we should not expose a counter, if user only use gauge's `inc` method, it act like a counter , if user change their mind to `dec` something, then it act like a gauge.

lust4life: but if prometheus treat them as different, then a `counter` can be simply wrapped by a interface only expose `inc` method, and check the value to be positive.

src/Logary/Metric/Gauge.fs
```diff
@@ -0,0 +1,79 @@
+namespace Logary.Metric
+
+
+type GaugeConf =
+  {
+   basicInfo: BasicConf
+   histogramConf: HistogramConf option
+  }
+
+  interface MetricBuilder<IGauge> with
```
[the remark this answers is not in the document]
lust4life: IGauge is something like 
https://godoc.org/github.com/prometheus/client_golang/prometheus#Gauge 

`MetricBuilder<IGauge>` here means this conf can produce a `IGauge` metric, this can give one conf to produce different `IMetric` , if their conf are same. (e.g. `counter`, `gauge`) , which will be use depend on the type inference

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+
+
+  testCase "histogram with lable" <| fun () ->
+    let testBuckets = [| 5.; 10.; 50.; 100.; |]
+    let registry = new MetricRegistry()
+
+    let histogramConf = HistogramConf.create {name = "some histogram"; description = "some histogram"; labelNames = [| "queue name" |]} |> HistogramConf.buckets testBuckets
+    let histogram = histogramConf |> registry.registerMetric
+
+    (histogram.labels [| "queue a" |]).observe 15.
```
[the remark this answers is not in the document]
lust4life: `Time` here can be express as `Span` in logary. 
```f#
    let lgb = logm.getLogger "logger.b"
    use _ = lgb |> Span.create |> Span.setMessage (eventX "some child span") |> Span.start
    // do actual work.
``` 
after span finished(disposed), it will have a timespan info which will be used for gauges.

src/Logary.Tests/Metrics.fs
```diff
@@ -0,0 +1,168 @@
...
+    gauge.set 200.
+    gauge.dec 140. // 60
+    gauge.dec 15.  // 45
+    gauge.dec 5.   // 40
+    gauge.dec 15.  // 25
+    gauge.dec 10.  // 15
+    gauge.dec 8.   // 7
+    gauge.inc 18.  // 25
+
+    let metricInfos = registry.getMetrictInfos ()
```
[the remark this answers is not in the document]
lust4life: get it .

src/Logary.Prometheus/Exporter.fs
lust4life: see explains below

src/Logary.Prometheus/Exporter.fs
lust4life: 👍

src/Logary/Extensions/LogManager.fs
lust4life:
```
  let enableCounterMetric (counterConf: Metric.BasicConf) (message: Message) =
    message |> setContext KnownLiterals.CounterMetricConfContextName counterConf
```
In fact, I don't recommend such a way to bind messages(prefer **remove this function**),  one should add a metric type separately instead of using this shortcut. Then no ship configuration in the log message.

src/Logary/MessageModule.fs
lust4life: In fact, I don't recommend such a way (`enableCounterMetric`) to bind messages(prefer remove this function), one should add a metric type separately instead of using this shortcut. Then no ship configuration in the log message.

i think this may some misunderstand here, "global style" it means use the global `DefaultMetricRegistry` in `LogManager`, but if user can grab a `LogManager` instance, then they can use logManager's `metricRegistry` to register these metrics (not the global one, the one in logManager's instance).

and the `enableCounterMetric` style is used with `logManager's instance style`, since metric conf is bind with a log message in this way, a log message will be send through a logManager(logary `Registry`), at that time it hook up the `metricRegistry` at that logary's registry instance. **This just save time for define another separate metric**. but this will `register/find` the metric on the fly (when sending log message), **which is not recommend** through https://prometheus.io/docs/instrumenting/writing_clientlibs/#metrics.

src/Logary/Global.fs
lust4life: #405

lust4life:
> We have IoT metrics being fed into Rutta, three instances of them. Each message has a value, that corresponds to a "message type" in a regular messaging system. Some messages are interesting; we'd like to create a histogram based on e.g. message type "scanned QR"-valued messages' field: "vehicle state", which can then be fed as Message-values into e.g. Google PubSub. I would have liked to write it with the Flow feature, but I suppose that it would in part intersect with your new histograms here. How would you do it?

> Also, would you do it differently when Rutta is replicated x3 or even x10, in order to handle the load? The histogram would be similar but slightly different between the replicas, which could be just fine.

I haven't try with Rutta, generally, if we use prometheus, it suggests to metric value or aggregate on each local machine (using metric api directly on each machine, expose them to prometheus's config), when prometheus pull these value, export them as a text protocol. 

But from your description above, it seems that these metric express as messages, needs aggregate on rutta, I think we can find out these message which we want to create metric with them, then use `Metric api in this pr` in rutta to metric them (but this will need deal with message's timestamp, since these metric are not occur when message occur, message may reached with delay). Then export these metric at rutta's level, which expose to prometheus.

Means that each rutta have its own global `MetricRegistry`. and since message will not duplicate on rutta no matter how many replications it will have. so i thins this will not be a problem

lust4life:
but I think the difficult part is,how to deside the metric name or labels or buckets only through the message,if users have no scenes about the metric,this may a little difficult

src/Logary.Tests/Metrics.fs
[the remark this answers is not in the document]
haf: True

haf:
> it suggests to metric value or aggregate on each local machine (using metric api directly on each machine, expose them to prometheus's config), when prometheus pull these value, export them as a text protocol.

Absolutely, that is done excellently here in your PR!

But these metrics would have tags, like what pod, service, namespace, rack, zone, region, etc, is displaying the metric, so sending them to an aggregator like Prometheus or Rutta wouldn't lose information (necessarily)

One issue that has killed our prometheus the last months is when programmers accidentally export label values to prometheus containing user-provided input or corresponding high-cardinality values. You got a nice function for displaying metrics/histograms here (!) — do you think we should add something to protect, if only ever so slightly, against high-cardinality label sets?

haf:
Mono is broken on Travis again :(

image

lust4life:
> One issue that has killed our prometheus the last months is when programmers accidentally export label values to prometheus containing user-provided input or corresponding high-cardinality values. You got a nice function for displaying metrics/histograms here (!) — do you think we should add something to protect, if only ever so slightly, against high-cardinality label sets?

Yes, nice suggestion, Although metric could on multi-machines, if one machines check they are high-cardinality, can throw a warning for users.

lust4life:
ci in travis is a little bit strange, re-run some old success build, make them failed to install mono 5.18.0 again, like below: 
```
Installing Mono
E: Unable to locate package referenceassemblies-pcl
The command "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget referenceassemblies-pcl" failed and exited with 100 during .
```

Post a topic in travis-ci's community forum: https://travis-ci.community/t/unable-to-locate-package-referenceassemblies-pcl/2686

or should we use mono `5.18.1` as it is the latest Stable Mono release.

src/Logary.Tests/Metrics.fs
```diff
@@ -175,4 +180,14 @@ let tests = [
...
     let histogramConf = HistogramConf.create("some histogram", "some histogram") |> HistogramConf.exponentialBuckets 5. 5. 5
     histogramConf.buckets |> Expect.sequenceEqual "should have buckets" [ 5.; 25.; 125.; 625.; 3125.; ]
+
+  testCase "avoid high cardinality" <| fun () ->
+    let registry = new MetricRegistry()
+    let gauge = {basicConf with labelNames = [| "user_id" |]} |> GaugeConf.create |> registry.registerMetric
+    for userId in 1..150 do
+      gauge.labels [| string userId |] |> ignore
+
+    Expect.throws "should throw if reach the high cardinality limit" (fun () ->
```
haf: Can we make it non-throwable by default, perhaps? That way we never throw exceptions, and we have to "opt in" to throwing them. So by default, we should perhaps just warn if we have a high cardinality data set (with internal loggar at warn) and then throw if the user opts into that.

lust4life: how about this ?

lust4life: non-throwable by default, needs logger to be passed in, do you have any ideas how to make it default since metric can be created outside of logary's registry (which have its runtime info with internal logger).

lust4life: 👍

haf:
Let's see if this fixes the build.

haf:
@lust4life I've added a section in the README so people know how to configure a metric with Prometheus. Perhaps you could add one of the samples from this thread there?

build.fsx
```diff
@@ -145,7 +138,7 @@ Target.create "Build" <| fun _ ->
     { MSBuild.CliArguments.Create() with
         Verbosity = Some Quiet
         NoLogo = true
-        Properties = [ "Optimize", "true"; "DebugSymbols", "true" ] }
+        Properties = [ "Optimize", "true"; "DebugSymbols", "true"; "Version", release.NugetVersion ] }
```
lust4life: Nice 👍

build.fsx
```diff
@@ -138,7 +138,7 @@ Target.create "Build" <| fun _ ->
     { MSBuild.CliArguments.Create() with
         Verbosity = Some Quiet
         NoLogo = true
-        Properties = [ "Optimize", "true"; "DebugSymbols", "true"; "Version", release.NugetVersion ] }
+        Properties = [ "Optimize", "true"; "DebugSymbols", "true"; "Version", release.AssemblyVersion ] }
```
lust4life: Nice 👍

haf: 😌

haf:
I've asked Travis engineers to have a look at the failing build

haf:
I've asked Travis engineers to have a look at the failing build

haf:
@lust4life Congratulations to completing this major feature to you (and us!) 🎊

jkone27:
how would i use Logary to export to prometheus in F# with suave? can i find docs for that?
logary.tech seems to be offline / unreacheable

PR #396, opened 2019-02-24
Feature/dash

haf:
screenshot 2019-02-24 at 19 22 44

 - nice to be able to show ongoing ratio of a log field -> calculate metric from logs
 - ability to do larger range of filtering than circular buffer; adjustable via UI
 - sending "filter at source" to backend


haf:
So the dash is completed, but I haven't released an app with it yet and I haven't decided how to authenticate users yet either... /cc @lust4life

PR #439, opened 2020-03-08
Add source link configuration.

dam5s:
This would help browsing the code during development and potentially during debugging sessions too.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,0 +1,12 @@
+<Project>
+  <PropertyGroup>
+    <!-- SourceLink related properties https://github.com/dotnet/SourceLink#using-sourcelink -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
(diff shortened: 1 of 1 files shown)


haf:
Thank you for your PR; there are some funkyness when it comes to the linked submodules, like can be seen here https://ci.appveyor.com/project/haf/logary/builds/31325271#L1865

There are also some tests that are mysteriously failing, but I can fix those once the actual source link PR works. :)

haf:
Strange, can't reproduce failure locally...

dam5s:
Yeah, I'm not super familiar with the build, I was under the impression the changes added a bunch of warnings which is a bit annoying, but it seemed to work otherwise.

haf:
@dam5s I've finally got the Travis build passing and speedy, and now we seem to be at the point where this PR is actually being tested. I thank you for your patience meanwhile.

I added a NuGet.props (which is not in use yet); and I see that msbuild automatically picks up on your target/props file? Could you give me a brief into to how this PR is intended to work?

Thank you,
Henrik

haf:
Closing due to inactivity.

haf:
I rebased your PR as #450

PR #450, opened 2020-11-05
Add source link configuration.

haf:
Rebased #439


haf:
Thanks @dam5s — for your patience. I know myself, and I indeed have too little patience for debugging CI jobs. In any case, we're merged now.

Tell me, are you using master? Because I haven't released v6 of the code-base yet. What's your use case?

